PR #518, opened 2024-03-06
fix: fixed emcc to use Os instead of O0

twilson63:
Changed emscripten build back to Os - was set to O0 for debugging.


TillaTheHun0 (approved):
LGTM

PR #524, opened 2024-03-09
feat(cu): add support for `WALLET_FILE` env var

MichaelBuhler:


TillaTheHun0:
Thanks for contributing.

I like that you kept all the parsing in the server config, whose job is to encapsulate server environment details away from the domain logic. In other words, the domain logic still only knows about `WALLET` :thumbsup: .

The `if` at the end feels disjointed. I bet using [`preprocess`](https://github.com/colinhacks/zod?tab=readme-ov-file#preprocess), you could parse `WALLET_FILE` as part of the schema, while still having `WALLET` be`required`, and remove the disjointed `if`.

WDYT?

MichaelBuhler:
> I like that you kept all the parsing in the server config, whose job is to encapsulate server environment details away from the domain logic. In other words, the domain logic still only knows about WALLET 👍 .

@TillaTheHun0 I noticed that from your clear comment.

I pushed another commit using `z.preprocess()`. It's lengthy, but readable.

WDYT?

TillaTheHun0 (approved):
LGTM. Excellent

Issue #526, opened 2024-03-11
import objects from "@permaweb/aoconnect" get error in a Next.js project when using NextJS App Router


iamgamelover:
Hello everyone,

I used `import { createDataItemSigner, message, dryrun } from "@permaweb/aoconnect";` in a Next.js project.
Running on localhost, get an error on the browser page.
```
./node_modules/@permaweb/aoconnect/dist/index.js
Attempted import error: 'warp-arbundles' does not contain a default export (imported as 'WarpArBundles').
```
I changed to `const { createDataItemSigner, message, dryrun } = require("@permaweb/aoconnect");` and it's working.
But when I'm deploying it to vercel.com, get a new error on the vercel deploy page:
```
TypeError: Cannot assign to read only property 'name' of function 'class A extends Error{static{this.name="InvalidSchedulerLocation"}constructor(...e){super(...e),this.name=A.name}}'
--
 at <static_initializer> (/vercel/path0/.next/server/app/testao/page.js:34:124266)
at 25290 (/vercel/path0/.next/server/app/testao/page.js:34:124244)
at t (/vercel/path0/.next/server/webpack-runtime.js:1:143)
at 8672 (/vercel/path0/.next/server/app/testao/page.js:69:2159)
at t (/vercel/path0/.next/server/webpack-runtime.js:1:143)
```
Anybody can give me some advices? Thanks.

TillaTheHun0:
Hey @iamgamelover , sorry you're having issues.

Both of these issues are downstream from NextJS' config. The first issue is most likely due to NextJS webpack config -- `aoconnect` is not importing a default export from `warp-arbundles`. This is the line in `aoconnect/dist/index.js`:

```js
import * as WarpArBundles from "warp-arbundles";
var { createData, ArweaveSigner } = WarpArBundles;
```

I suspect something with NextJS webpack config is trying to support the myriad of ESM+CJS combinations that exist in NodeJS ecosystem, and is getting tongue tied. Sounds like the `require` get's around that issue though.

The seconds issue is an error in NextJS' static code analysis (also being done by Webpack, it seems). It's claiming that read-only property `name` cannot be reassigned, in relation to [this error class](https://github.com/permaweb/ao/blob/a6d0d1bb4a2d868002b9e5da2a517127b65aef4a/scheduler-utils/src/err.js#L1)

[...]

PR #527, opened 2024-03-11
`aoconnect` to support for NextJS apps

kernelwhisperer:
Will add a description shortly.


TillaTheHun0:
Thanks. This is to do with NextJS' Webpack config (see https://github.com/permaweb/ao/issues/526#issuecomment-1988504950), but if this get's around those issues, then we might as well make the change on our end.

No minifying stinks for vanilla web apps importing the module directly from the browser, but can revisit another day.

TillaTheHun0 (approved):
Thanks!

PR #540, opened 2024-03-14
feat(ur): add `redirect` strategy

### LINKED ISSUE #539: feat(ur): add ability to redirect instead of rev proxy
Right now the Unit Router rev proxies to the underlying hosts. We would like to add the ability to configure it to redirect instead of rev proxy.

Once the application code changes are made, we will need to leverage it to redirect instead of rev-proxy to the MUs and CUs running testnet. This will require also setting up SSL on each box by way of an nginx server on each box that rev-proxies to the application on localhost. Similar work was done on the [SUs here](https://github.com/permaweb/ao-infra/pull/22). This can most likely be lifted into MUs and CUs `userdata` and infra.

## Acceptance Criteria

- We can set an environment variable to set the router in "redirect" or "rev-proxy" mode. The environment variable should default to "rev-proxy", for backwards compat. Later on, once we've verified "redirect" is working, we can set it as the default.

- The status code should be a `307` which indicates to the client to resend the body as well, useful for `POST` requests that are made to the Messenger Unit.

- MUs and CUs can be consumed directly over `https` ie `https://cu1.ao-testnet.xyz`

- Unit router is deployed to redirect traffic to MUs using `https`
- Unit router is deployed to redirect traffic to CUs using `https`

## Developer Notes

- We will need to update the `HOSTS` environment variable to use `https` [here](https://github.com/permaweb/ao-infra/blob/c688cef28e047b91cb237552548e45e99dd1ff9a/terraform/src/modules.tf#L1)

- As an added optimization we may want to add in `aoconnect` the ability to cache the redirect, following it manually once, then caching the returned location. This reduces a roundtrip, similar to what [`scheduler-utils` can be configured to do](https://github.com/permaweb/ao/blob/6fc7052e13766682ad6159990e2365d0f2dfc5a2/scheduler-utils/src/index.common.js#L16) (this should be a separate issue, if done, just highlighting here as related.)

MichaelBuhler:
Adds `STRATEGY` environment variable, with default value of `proxy`. This maintains the current behavior.

With `STRATEGY=redirect`, the UR to return HTTP 307 redirects instead.

Closes #539.

### DIFF
--- a/servers/ur/src/app.js
+++ b/servers/ur/src/app.js
@@ -1,84 +1,43 @@
-import { Readable } from 'node:stream'
+import { join } from 'node:path'
+
+import cors from 'cors'
 import express from 'express'
-import WarpArBundles from 'warp-arbundles'
+import heapdump from 'heapdump'
+import { pipe } from 'ramda'
 
-import { router } from './router.js'
+import { config } from './config.js'
+import { logger } from './logger.js'
 
-const { DataItem } = WarpArBundles
+import { proxyWith } from './proxy.js'
+import { redirectWith } from './redirect.js'
 
-/**
- * The ONLY custom bits needed for the router.
- *
- * Mount any custom endpoints, optionally reverse-proxying to the set of
- * underyling hosts with automatic failover, by using the injected revProxy handler
- */
-
-/**
- * The Reverse Proxy Configuration for an ao Compute Unit Router
- */
-function aoComputeUnitMount ({ app, revProxy }) {
-  app.get('/', revProxy({ processIdFromRequest: (req) => 'process' }))
-  app.get('/result/:messageTxId', revProxy({ processIdFromRequest: (req) => req.query['process-id'] }))
-  app.get('/results/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.get('/state/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.get('/cron/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.post('/dry-run', revProxy({ processIdFromRequest: (req) => req.query['process-id'] }))
+const middlewareWithByStrategy = {
+  proxy: proxyWith,
+  redirect: redirectWith
 }
 
-/**
- * The Reverse Proxy Configuration for an ao Messenger Unit Router
- */
-function aoMessengerUnitMount ({ app, revProxy }) {
-  class InvalidDataItemError extends Error {
-    constructor (...args) {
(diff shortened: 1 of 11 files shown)


servers/ur/src/app.js
```diff
@@ -1,84 +1,43 @@
...
-function aoComputeUnitMount ({ app, revProxy }) {
-  app.get('/', revProxy({ processIdFromRequest: (req) => 'process' }))
-  app.get('/result/:messageTxId', revProxy({ processIdFromRequest: (req) => req.query['process-id'] }))
-  app.get('/results/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.get('/state/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.get('/cron/:processId', revProxy({ processIdFromRequest: (req) => req.params.processId }))
-  app.post('/dry-run', revProxy({ processIdFromRequest: (req) => req.query['process-id'] }))
+const middlewareWithByStrategy = {
+  proxy: proxyWith,
+  redirect: redirectWith,
```
TillaTheHun0, line 16: Dangling commas should be getting caught by the linter `standard` which runs on pre-commit. How's this getting through?

servers/ur/src/redirect.js
```diff
@@ -0,0 +1,37 @@
...
+   * by the injected business logic.
+   */
+  const redirectHandler = ({ processIdFromRequest }) => {
+    return async (req, res) => {
+      const processId = await processIdFromRequest(req)
+      const host = determineHost({ processId })
+
+      _logger('Redirecting process %s to host %s', processId, host)
+
+      const location = new URL(req.originalUrl, host)
```
TillaTheHun0, line 26: I'd add a comment linking to express documentation here: https://expressjs.com/en/api.html#req.originalUrl

TillaTheHun0 (commented):
Looks great!

For some reason, standard isn't fixing your dangling commas, which it should on `pre-commit`.

Also one suggestion for posterity's sake.

TillaTheHun0:
I'd also link your commit to the issue, when possible:

```
feat(ur): add redirect strategy #539
```

servers/ur/src/redirect.js
MichaelBuhler: Will do. I still have the tab open 😅

servers/ur/src/app.js
MichaelBuhler: I guess I never ran `npm install` at the monorepo root. Will fix.

TillaTheHun0 (approved):
LGTM

PR #557, opened 2024-03-21
feat(cu): implement Assignment type handling #556

### LINKED ISSUE #556: feat(cu): implement Assignment protocol spec
When transforming a page of messages retrieved from a Scheduler Unit, for a given process, if the CU comes across an `Assignment` in the list, it should:

- transform only the metadata fields for the assignment:
  - `name`
  - `ordinate`
  - `Id`
  - `From`
  - `Forwarded-By`
  - `Epoch`
  - `Nonce`
  - `Timestamp`
  - `Block-Height`
  - `Hash-Chain`
- If the assignment references a `Message` -- a tx id on-chain, then the CU should "hydrate" the assignment with the transaction referenced by the `Assignment`, constructing a bonafide message to be passed to the process:
  - `Id`
  - `Tags`
  - `Data`
  - `Signature`
  - `Anchor`
  - `Owner`
  - `From` (set to `Owner`)

## Developer Notes

The tx id is attested to be on chain by the SU, as part of scheduling the `Assignment`.

The CU should still support `Load` messages, but we should make the changes necessary to "deactivate" `Load` messages at a particular block height. This is effectively deprecating `Load` messages

~If no `Message` is present on the `Assignment`, then the values for `Id`, `Tags`, `Data` `Signature`, `Anchor` and `Owner` will be set to the values from the `Assignment`, itself.~

## Questions

~#### Which field on the messages returned from the SU will contain the `Message` value on the `Assignment`, if it exists? I propose three potential impls:~

~1. set `node.message` to the tx id, if `Message` tag is provided. The CU would then need to distinguish object (regular message) vs string (assignment) vs none (empty assignment?) and act accordingly.~

~> downside of this approach is that we lose visibility into any info on the Assignment data item itself, ie. its own `tags`, `signature`, `anchor`, and `id`~

~2. set `node.assignment` to the tx id, if `Message` tag is provided. `node.message` would still contain the `Assignment` data item itself. The CU would then need to distinguish the presence of `node.assignment` (assignment) or its absence (regular message)~

~> downside of this approach is that we cannot distinguish between a "regular" message and an `Assignment` with no `Message` value set (which is valid according to the spec)~

~3. set `node.type` to `assignment` and `node.assignment` to the tx id, if `Message` tag is provided~

~#### Is there any functional difference between a regular message and an assignment with no `Message` tag specified?~

~#### Should `Owner` be set to the owner of the assignment or the owner of the `Message` pointed to, on chain? What about `From`?~

TillaTheHun0:
Closes #556 

Also puts in place a strategy to deprecate `Load` messages, supporting them until a TBD block height.

### DIFF
--- a/servers/cu/src/domain/index.js
+++ b/servers/cu/src/domain/index.js
@@ -201,6 +201,14 @@ export const createApis = async (ctx) => {
     locateProcess: locateDataloader.load.bind(locateDataloader),
     doesExceedModuleMaxMemory: WasmClient.doesExceedModuleMaxMemoryWith({ PROCESS_WASM_MEMORY_MAX_LIMIT: ctx.PROCESS_WASM_MEMORY_MAX_LIMIT }),
     doesExceedModuleMaxCompute: WasmClient.doesExceedModuleMaxComputeWith({ PROCESS_WASM_COMPUTE_MAX_LIMIT: ctx.PROCESS_WASM_COMPUTE_MAX_LIMIT }),
+    /**
+     * This is not configurable, as Load message support will be dictated by the protocol,
+     * which proposes Assignments as a more flexible and extensible solution that also includes
+     * Load Message use-cases
+     *
+     * But for now, supporting Load messages in perpetuity.
+     */
+    AO_LOAD_MAX_BLOCK: Number.MAX_SAFE_INTEGER,
     logger
   })
   /**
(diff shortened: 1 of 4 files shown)


TillaTheHun0:
After syncing with @VinceJuliano , it makes most sense for `message` on each `node` to be either an object _or_ a string, and then the CU to use that to discern

PR #586, opened 2024-04-03
chore(cu): replace PouchDB with Sqlite

### LINKED ISSUE #413: refactor(cu): swap out PouchDB for SQLite
Now that we won't be needing to persist Blobs like WASM Modules and Heaps (see #404 and #411), we do not need a persistence engine that supports it. For the sake of compatibility and ease of use, we can just use SQLite to persist evaluation results and metadata, but without the WASM heap.

Since we leverage ports and adapters, we will just be able to implement the DB api and none of the business logic will need to change.

Before we do this though, we must first do #404 and #411

## Acceptance Criteria

- [x] All previous tests pass. You may need to change dep injection, assertions in injected dependencies (to ensure correct parameters/querys are being made), and some scaffolding, but the output assertions and contract schema wrapping (`{zodSchema}.implement(...)`)should remain the same
- [x] PouchDB is removed as a dependency and all Pouch dependencies are removed from the manifest
- [x] README is updated to reflect changes including setup directions
- [x] Configuration + config model is updated to reflect any added (or removed) configuration required by the CU 
- [x] The CU behavior and public api is same as before. In other words, public api outputs between CU+PouchDB and CU+SQLite should be identical

## Stretch Goals

- [x] Ideally, the CU can still be spun in a turn-key way, ie. by only specifying `WALLET/WALLET_FILE` and `NODE_CONFIG_ENV`. To this end, having defaults for any new configuration required is desirable.

## Developer Notes

The current PouchDB is ephemeral, and so no data migration from PouchDB -> SQLite will be warranted.

There are currently 4 document types persisted into PouchDB:

- [x]  [`Block`](https://github.com/permaweb/ao/blob/4284805144585021a4d0da14f9c0b262e4d02ef5/servers/cu/src/domain/client/ao-block.js#L8) - represents a block and it's associated height and timestamp. Cached in persistence for when calculating Cron messages, so that each message has accurate Block height and timestamp metadata.

- [x]  [`Process`](https://github.com/permaweb/ao/blob/4284805144585021a4d0da14f9c0b262e4d02ef5/servers/cu/src/domain/client/ao-process.js#L127): represents a process and it's associated metadata ie. `tags`, `block`, etc. Cached in persistence to reduce calls to the Scheduler Units, as well as gateways to locate processes on a Scheduler Unit

- [x]  [`Evaluation`](https://github.com/permaweb/ao/blob/4284805144585021a4d0da14f9c0b262e4d02ef5/servers/cu/src/domain/client/ao-evaluation.js#L9): represents the result of a message evaluation. Cached so that previous evaluations can be served on the `/result` and /`results` endpoint

- [x]  [`Module`](https://github.com/permaweb/ao/blob/4284805144585021a4d0da14f9c0b262e4d02ef5/servers/cu/src/domain/client/ao-module.js#L9): represents a module and it's associated metadata ie. `tags`, `owner`, etc. Cached in persistence, so that each Module's tags can be used to verify CU support ie. `Module-Format` and instantiating the runtime constraints ie. `Memory-Limit` and `Compute-Limit`. Caching Module also allows for more quickly locating and downloading from Arweave.

Each of these documents are mapped to/from by domain models found in `model.js`, used in the business logic. Those _should_ not need to change, but it's not a strict requirement; there could be something i'm not taking into account that requires a change to one of those business models.

Each of these documents are written once to database, then never updated. In other words, persistence is append only and documents never change once created.

Each document has pretty clearly defined access and creation use-cases, as described by client contracts in `dal.js`, and also pretty clearly defined querying and creation logic, as implemented in each client. They are read often, so read/query efficiency is top of mind for the implementation. PouchDB uses a primary index and map reduce views to achieve querying efficiency, but SQLite impl doesn't need to follow that same approach -- developer's choice.

Subsequently, the changes that ought to be required should be almost entirely contained in the `client` folder and then wiring of dependencies in `domain/index.js`. Contracts in `dal.js` should not need to be updated, which will ensure the new implementations based on SQLite satisfy the business logic contracts.

There may be some missing tests ie. `findBlocksWith` and `saveBlocksWith`, and these ought to be implemented now, as part of this task (could be broken out into separate issues), now that we are revisiting this work and will need a way to assert correctness. There shape can be copied from an existing test ie. `saveProcessWith` or  `findEvaluationsWith`. I would recommend first implementing these tests to pass on the old impl, then update the implementation and dep injection in tests in order to pass the same assertions.

TillaTheHun0:
Closes #413 

No additional features. This PR replicates current functionality, but uses SQLite instead of PouchDB


twilson63 (approved):
LGTM

PR #587, opened 2024-04-04
add RESTRICT_TO_PROCESSES env var

smowden:
This PR introduces the possibility to whitelist only specific process ids.
The idea is to have read only CU's that server compute heavy processes.
The pinning of these processes can already be effecitvely achieved (i think) by setting limits such as PROCESS_MEMORY_CACHE_TTL, PROCESS_MEMORY_CACHE_MAX_SIZE very high. However you would not want someone else to get to comfy on your beefy CU in this scenario hence the allow list.


twilson63:
Talked with the team and currently can't see any holes with this approach,
I think we want to make sure we return a proper http message when the user
requests `/result` `/results` and `/cron` for a given process id, that the
CU returns a proper http status code and JSON message indicating that this
CU does not serve this process.

twilson63:
@TillaTheHun0 when you have time, can you look at this PR?

servers/cu/src/domain/model.js
```diff
@@ -128,7 +128,11 @@ export const domainConfigSchema = z.object({
    * The amount of time in milliseconds, the CU should wait for evaluation to complete
    * before responding with a "busy" message to the client
    */
-  BUSY_THRESHOLD: positiveIntSchema
+  BUSY_THRESHOLD: positiveIntSchema,
+  /**
+   * A list of process names that the CU should restrict to
+   */
+  RESTRICT_TO_PROCESSES: z.array(z.string())
```
TillaTheHun0, line 135: You can use the `commaDelimitedArraySchema` which will handle parsing the string and trimming any whitespace, or nooping if the value is already an array.

servers/cu/src/config.js
```diff
@@ -144,7 +145,8 @@ const CONFIG_ENVS = {
     PROCESS_CHECKPOINT_FILE_DIRECTORY: process.env.PROCESS_CHECKPOINT_FILE_DIRECTORY || tmpdir(),
     PROCESS_MEMORY_CACHE_MAX_SIZE: process.env.PROCESS_MEMORY_CACHE_MAX_SIZE || 500_000_000, // 500MB
     PROCESS_MEMORY_CACHE_TTL: process.env.PROCESS_MEMORY_CACHE_TTL || ms('24h'),
-    BUSY_THRESHOLD: process.env.BUSY_THRESHOLD || 0 // disabled
+    BUSY_THRESHOLD: process.env.BUSY_THRESHOLD || 0, // disabled
+    RESTRICT_TO_PROCESSES: process.env.RESTRICT_TO_PROCESSES ? process.env.RESTRICT_TO_PROCESSES.split(',') : []
```
TillaTheHun0, line 149: By using `commaDelimitedArraySchema` in the `domainConfigSchema`, this guard and parsing can be pushed down into the domain.

Then this line can just be:
```
RESTRICT_TO_PROCESSES: process.env.RESTRICT_TO_PROCESSES || []
```

servers/cu/src/routes/middleware/withProcessRestriction.js
```diff
@@ -0,0 +1,12 @@
+export const withProcessRestriction = (handler) => (req, res, next) => {
+  const {
+    params: { processId },
```
TillaTheHun0, line 3: Not all endpoints will have `processId` as a resource in the path ie. `.../:processId`. Some endpoints, like `/dry-run`, receive `process-id` as a query parameter.

Here's a exhaustive list of CU endpoints and where to retrieve the `processId`, for each

- `/cron/:processId` (path)
- `/dry-run?process-id=...` (query param)
- `/result/:messageId?process-id=...` (query param)
- `/results/:processId` (path)
- `/state/:processId` (path)

So you could consider creating a middleware that can dynamically extract the `processId` from the `req`, say a HoF, using an "`extractor`" injected as a parameter:

```js
const withProcessRestrictionFrom = ({ extractor }) => (handler) => (req, res, next) => {
  const processId = extractor(req)

  if (...) return res.status(403).send(...)

  return handler(req, res, next)
}

const withProcessRestrictionFromPath = withProcessRestrictionFrom({ extractor: (req) => req.params.processId })
const withProcessRestrictionFromQuery = withProcessRestrictionFrom({ extractor: (req) => req.query['process-id'] })
```

Then you could simply compose the appropriate middleware, depending on the route, instead of as part of `withMiddleware`

I hope that description makes sense. Other valid approaches as well.

TillaTheHun0 (commented):
Thank you for the PR! It's clear that you took time to get familiar with the overall project structure, and align your impl with that. We really appreciate that! :)

This looks great and seems nearly there. I have some suggestions/corrections for the impl. Please see my comments.

There's a [section in the `README`](https://github.com/permaweb/ao/tree/main/servers/cu#environment-variables) that details all environment variables, their purpose, and their default value (if one exists). Could you add this environment variable to that section in the `README`?

Finally, I notice your commits are not following [Conventional Commits Style](https://www.conventionalcommits.org/en/v1.0.0/#summary), so if you could please adjust those commit messages as well. There's some guidance in our [`CONTRIBUTING`](https://github.com/permaweb/ao/blob/main/CONTRIBUTING.md#commit-messages) file. You can also run `npm i` at the root of the repo to install git hooks that will automatically do this when you commit:
- lint your commit messages
- lint and format your code changes
- update ToCs in any Markdown files changed.

PR #595, opened 2024-04-08
Fix scheduler utils locate.js file's locateWith function

jsoneaday:
The locateWith function returns a specific type that looks like this { url, address }, as shown in lines 16 and 48. This object is also the only type used to save data into the cache in the in-memory.js file. Now, on line 27 it appears that an early return and exit of the function is being attempted. However this return call does not immediately exit the function. Instead it goes to the subsequent `next` call.

This is an issue because in this subsequent `next` call the object is being received as a type { url, ttl, owner }, which would only be true if the scheduler object came from one of the other returns. But in this case it came from the line 27 return which is of type { url, address }.

In addition to this issue in the function the last test in locate.test.js is also making a similar mistake. On line 158 getByOwner is returning a { url, ttl, owner } type. This is wrong since again the in-memory cache only saves the type { url, address }. Also line 164 needs an update as ttl is, again, not part of the cached object. And then finally the assert on line 177 should use DOMAIN and not DOMAIN_REDIRECT. Since if the function was to exit early on line 27 it would never hit the call to set finalUrl, which means that the url would never be reset to DOMAIN_REDIRECT.

My code changes have been tested to successfully run all tests inside of scheduler-utils.


TillaTheHun0:
Thanks for submitting the PR! Really awesome debugging, you're right that `ttl` is not present on the cached `byOwner` value, and so that branch will result in passing `undefined` for `ttl` when `setByProcess` is invoked.

The solution in the PR would mean that `locate` would no longer cache location by process, when using a `schedulerHint`, which isn't great. I think I would prefer to instead just store `ttl` in the cache entry in `setByOwnerWith`. Since `raw` only returns `{ url }` and `validate` returns a `boolean`, this wouldn't expand the public api surface, but would ensure `ttl` is available on all branches that lead to line 47 where `setByProcess` is passed a `ttl`.

I hope I explained that well.

Some other things:

- The JavaScript in the repo follows [Standard style](https://standardjs.com/). You can run `npm i` at the root of the repo to install git hooks to automatically check and format JS code in your commit, when you commit.
- There looks to be a lot of extra commits from some other work you're doing (looks like you're working on Rust-based CU -- cool!), but that was all deleted in `613974b0ede2937410c4bc79c063708780371fd4`. It would be best not to have that in the ao main repo history, for this fix. Maybe consider branching off `main` and implementing the `scheduler-utils` fix there, so that the PR to here only contains that fix.

jsoneaday:
Yeah sorry about all the rust stuff. I only saw that issue once I had already done a ton of rust related items.

Yeah I wasn't entirely sure my fix was the right thing to do because the intention wasn't entirely clear to me. I was only trying to be consistent with what I saw on line 16 and didn't want multiple types being used.

I'm a bit confused by your second paragraph. When line 27 runs it skips line 30 and just goes straight to the next then statement. So when byOwner comes into the subsequent next statement it does not have a ttl since the cache.setByOwner call never ran (if it had run the assert.fail would have triggered which it does not). Also I personally prefer being very strict about type schema. I understand that the addition of ttl won't really be seen elsewhere, but since js is not statically typed it may lead to more confusion down the road. 

I did try and run npm i on the root of the project, but got this error: (base) davidchoi@Davids-MacBook-Pro-2 ao % npm i
npm ERR! code EEXIST
npm ERR! syscall rename
npm ERR! path /Users/davidchoi/.npm/_cacache/tmp/4d3e4f56
npm ERR! dest /Users/davidchoi/.npm/_cacache/content-v2/sha512/bc/16/b5783f533a6aeed7c2c1896abfb36ba78a1bdf6d1e138f96489b0801df4fe11c8d64431617ae2d2fb6d348f2b0c69b4065fcb064300117e0f8b33b8e891e
npm ERR! errno -13
npm ERR! EACCES: permission denied, rename '/Users/davidchoi/.npm/_cacache/tmp/4d3e4f56' -> '/Users/davidchoi/.npm/_cacache/content-v2/sha512/bc/16/b5783f533a6aeed7c2c1896abfb36ba78a1bdf6d1e138f96489b0801df4fe11c8d64431617ae2d2fb6d348f2b0c69b4065fcb064300117e0f8b33b8e891e'
npm ERR! File exists: /Users/davidchoi/.npm/_cacache/content-v2/sha512/bc/16/b5783f533a6aeed7c2c1896abfb36ba78a1bdf6d1e138f96489b0801df4fe11c8d64431617ae2d2fb6d348f2b0c69b4065fcb064300117e0f8b33b8e891e
npm ERR! Remove the existing file and try again, or run npm
npm ERR! with --force to overwrite files recklessly.

npm ERR! A complete log of this run can be found in: /Users/davidchoi/.npm/_logs/2024-04-08T23_18_57_817Z-debug-0.log.

I guess you're asking me to put in a new pr that is a fresh download of main, without my rust stuff? If you could clarify the second paragraph I should be able to implement it.

TillaTheHun0:
That `npm` error looks to be something with your local environment -- an `npm` cache file not able to be deleted or something. Maybe consider running removing that file or running `npm i --force` as in the suggestion. After that, it will probably work fine.

> I guess you're asking me to put in a new pr that is a fresh download of main

There's been changes to `scheduler-utils` since you forked, so may need to get those changes into your fork anyway. Since Git seems to have no conflicts with your current branch, you may just be able to instead just branch off the beginning of your fork, which wouldn't include all your rust code:

```
git checkout -b jsoneaday/scheduler-utils-ttl-byOwner e5fe11b254089017a0202dde2a60e8468d0caea8
```

Either way, once you have a copy without your rust changes, you would implement your changes there.

In terms of the impl:
- modify `setByOwnerWith` to also store `ttl` in its cache entry. That way, `ttl` is available on line 27 return of `locateWith`. - Will need to update test `getByOwnerWith.returns the url if in cache` to assert `ttl` is stored in the entry

Then _I think_ all the other impl could remain the same. Does that make sense?

jsoneaday:
I was able to get the root npm i to install by using sudo. 
```
(base) davidchoi@Davids-MacBook-Pro-2 ao % sudo npm i --force
Password:
npm WARN using --force Recommended protections disabled.

> prepare
> husky


added 480 packages, and audited 487 packages in 2s

156 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```
And I can create the new branch jsoneaday/scheduler-utils-ttl-byOwner as well.

The issue with your request is that the cache.getByOwner schema is { url, address }. So even after adding the ttl to the setByOwner cache, when line 46 gets called scheduler.owner will be undefined since the cache doesn't have an owner field it has an address field. I suppose on line 46 we could do something like this
`const byProcess = { url: finalUrl, address: scheduler.hasOwnProperty("owner") ? scheduler.owner : scheduler.address }`
Would that be ok? 

Also if we change the schema set by setByOwner to add the ttl I suggest we should do the same to setByProcess for consistency sake.

jsoneaday:
Ok so then to clarify, the change gives a unique schema for getByOwner and setByOwner to { url, ttl, owner } while getByProcess and setByProcess remain as { url, address }.
This change updates in-memory.js, locate.js and the locate.test.js files.

TillaTheHun0:
> getByOwner and setByOwner to { url, ttl, owner } while getByProcess and setByProcess remain as { url, address }.

That would also be fine, but not what I was thinking. Sorry that I wasn't clear. My suggestion is:

- add `ttl` to `byOwner` cache entries
- change line 27 in `locateWith` to map `byOwner` to need shape for `next`:
```js
if (byOwner) return { url: byOwner.url, owner: byOwner.address, ttl: byOwner.ttl }
```
- fix `locateWith.should use the scheduler hint and use the cached owner` test to return actual shape from `getByOwner`

Either what you suggested above, or my suggestion, would be good.

jsoneaday:
I submitted a new pr on the branch you gave me https://github.com/permaweb/ao/pull/603. I've used the method I suggested because I'm concerned about other code using setByOwner and expecting the prior schema.
I've seen something that makes me a bit concerned but let's have that discussion on the new pr

PR #600, opened 2024-04-09
fix(mu): use graphql_url instead of gateway_url

MichaelBuhler:
This change is brought about by changes in scheduler-utils@0.0.16.

And we're mimicing similar changes in the CU.


TillaTheHun0 (commented):
This looks right to me. It's definitely needed for `scheduler-utils`, but @VinceJuliano I defer to you on whether this works with rest of the MU -- idk if `GATEWAY_URL` is used elsewhere in the MU

VinceJuliano:
Ran this locally works for me

PR #615, opened 2024-04-12
feat(cu): add whitelist support

smowden:
adding whitelist support while following @TillaTheHun0's lead as close as possible

renamed the env var to ALLOW_PROCESSES because i thought RESTRICT_PROCESSES and RESTRICT_TO_PROCESSES would be very prone to user mix ups

Original PR:
https://github.com/permaweb/ao/pull/587


TillaTheHun0 (dismissed):
LGTM. Thanks!

servers/cu/src/routes/middleware/withProcessRestriction.js
```diff
@@ -1,15 +1,15 @@
...
-      // See https://github.com/permaweb/ao/pull/587
+      RESTRICT_PROCESSES,
+      ALLOW_PROCESSES
     }
   } = req
 
   const processId = extractor(req)
 
   if (RESTRICT_PROCESSES && RESTRICT_PROCESSES.includes(processId)) return res.status(403).send({ error: `Access denied for process ${processId}` })
+  if (ALLOW_PROCESSES && !ALLOW_PROCESSES.includes(processId)) return res.status(403).send({ error: `Access denied for process ${processId}` })
```
TillaTheHun0, line 12: Since this is a `whitelist`, I think this will need to also check `length`. Otherwise, I think an empty list will bar all processes.

smowden: you are absolutely right

TillaTheHun0 (approved):
LGTM!

PR #627, opened 2024-04-17
Feat: Multichain dataitem support

atticusofsparta:
Implements usage of arbundles DataItem in the MU

Updates SIG_CONFIG on the SU to support relavent chains (config taken from arbundles)

add node env signer functions for eth and sol. I did attempt browser versions but it needs alot more attention, so figured this should be added first... browser signing can be implemented externally anyways :)


connect/package.json
```diff
@@ -44,8 +44,13 @@
   },
   "dependencies": {
     "@permaweb/ao-scheduler-utils": "~0.0.16",
+    "@solana/web3.js": "^1.91.6",
+    "arbundles": "^0.11.0",
```
dtfiedler, line 48: nit: you may want to pin `arbundles`

zyjblockchain:
May I ask how long it will take for this PR to be approved?

PR #640, opened 2024-04-24
fix(connect): add missing limit arg to results api

femaury:
The docs + the ao-cu implementation talk about/use a `limit` property which was missing from the results api.
The zod object also had a property `limit`, although it was marked as a string. I've changed this to be a number.


TillaTheHun0 (approved):
LGTM. Thanks for the contribution!

PR #669, opened 2024-05-03
fix(cu): add `long-timeout` dependency, update cu

dtfiedler:
TODO: add tests with MAX_SAFE_INTEGER as TTL


dtfiedler:
Resolved in - https://github.com/permaweb/ao/pull/788

Sorry for the lack of follow up here!

PR #675, opened 2024-05-06
feat(cu): dry-run only uses latest memory for process to perform dry-run evals #674

### LINKED ISSUE #674: feat(cu): dry-run only to be performed using the latest previously evaluated process memory
## Background

Right now, `dry-run` will cause the CU to fetch any messages from the SU it needs in order to "catch up" to the place the dry-run operation is requesting to dry-run the provided message. This is typically `latest`.

## Problem

This causes messages to be evaluated in the CU, that ought be evaluated as part of a call to `/result`. This is not kosher with the ao Data Protocol.

## Solution

Dry Run should instead only perform the dry-run using the latest _previously_ evaluated process memory, according to the CU. This means new messages will be evaluated _only_ as part of `/result` or `/cron` from a MU monitor.

This is to say that the CU will use the latest process memory, either from the CUs LRU In-Memory Cache, a file checkpoint, or a Checkpoint queried from Arweave, to perform the dry-run. Dry Run will not result in the evaluation of _new_ messages.

TillaTheHun0:
Closes #674 

`readState` is no longer invoked by `dryRun`. Instead `dryRun` simply loads the process (which includes caching process, loading it's latest memory, and priming the LRU In-Memory cache if needed), loads the module (which includes caching),
then finally evaluates the single dryRun message

### DIFF
--- a/servers/cu/src/domain/api/dryRun.js
+++ b/servers/cu/src/domain/api/dryRun.js
@@ -1,162 +1,115 @@
 import { Readable } from 'node:stream'
 import { omit } from 'ramda'
-import { Resolved, of } from 'hyper-async'
+import { of } from 'hyper-async'
 
-import { loadMessageMetaWith } from '../lib/loadMessageMeta.js'
+import { loadProcessWith } from '../lib/loadProcess.js'
+import { loadModuleWith } from '../lib/loadModule.js'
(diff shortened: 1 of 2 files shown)


twilson63 (approved):
LGTM

PR #703, opened 2024-05-12
fix(ur): do not parse body on root mu POST route, if request is an assignment #697

TillaTheHun0:
This PR makes it so that the MU route will conditionally consume the body based on whether or not the `assign` query parameter is provided. In this case, there is no body to consume. Additionally, the `processId` is available as a `process-id` query parameter, and so can be extracted from there.


VinceJuliano (approved):
LGTM!

PR #727, opened 2024-05-22
chore(cu): enforce side effect schemas #721

### LINKED ISSUE #721: chore(cu): enforce each side effect with a schema
## Background

We use `zod` to enforce contracts between our business logic and side effect implementations. These contracts are leveraged by the business logic to [wrap any injected side effects](https://github.com/permaweb/ao/blob/c834e0ac5ff4aff5b26200b7f253b27568d640dd/servers/cu/src/domain/lib/loadProcess.js#L90). This ensures, not only, that our side effects implement the schema, but that our [stubs used in unit tests](https://github.com/permaweb/ao/blob/c834e0ac5ff4aff5b26200b7f253b27568d640dd/servers/cu/src/domain/lib/loadProcess.test.js#L41) also correctly implement the schema.

## Problem

For the sake of expediency, some side effects were injected without being wrapped by a Zod Schema. We need to revisit those instances, implement the schemas, then wrap the side effects with those schemas.

Some examples of this:

- [ ] `isProcessOwnerSupported` schema and wrapping (LOE low)
- [ ] `findLatestProcessMemory` schema and wrapping (LOE medium)
- [ ] `saveLatestProcessMemory` schema and wrapping (LOE medium)

## Developer Notes

Depending on the lift, it may be worth splitting any of the work for a single side effect into a separate issue.

jfrain99:
Closes #721 

Adds schema for side effects and tests for saveLatestProcessMemory.
Also fixes bug related to cache timestamp comparison for saving checkpoint.

### DIFF
--- a/servers/cu/src/domain/client/ao-process.js
+++ b/servers/cu/src/domain/client/ao-process.js
@@ -1242,7 +1242,7 @@ export function saveLatestProcessMemoryWith ({ cache, logger, saveCheckpoint, EA
      * Having updateAgeOnGet to true also renews the cache value's TTL,
      * which is what we want
      */
-    if (cached && isLaterThan(cached, { timestamp, ordinate, cron })) return
+    if (cached && !isLaterThan(cached.evaluation, { timestamp, ordinate, cron })) return
 
     logger(
       'Caching latest memory for process "%s" with parameters "%j"',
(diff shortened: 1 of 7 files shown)


TillaTheHun0 (approved):
LGTM. Well done! And awesome job implementing the tests for `saveLatestProcessMemory`!

PR #735, opened 2024-05-24
fix(mu): handle duplicate from-process tags on transactions #660

### LINKED ISSUE #660: fix(mu): duplicate From-Process tags on transactions
Either Messages coming in from users and coming out of process results have been reported to have duplicate From-Process tags there should only be one on each message. We should check both.

Tags in the MU are added in the build-tx.js function. It may be that these duplicates are coming from aoconnect though and not the MU.

jfrain99:
Closes #660 
MU is filtering out duplicates, but they are generated in ao-connect by users. Decided to be nonissue, but test added to ensure duplicate From-Process tags are outputted as one tag.


VinceJuliano (approved):
Looks great!

PR #736, opened 2024-05-24
feat(connect): add serializeCron util function to connect

### LINKED ISSUE #206: connect: add utility to serialize Cron "message" into flattened `Cron-Tag-{Name}` tags
```
As an aoconnect user
I would like a utility that will serialize any Cron data as tags on my process
So that I can more seamlessly spawn my Cron-utilizing process
```

## Acceptance Criteria

- [ ] function of the following shape is added to `aoconnect` and exported
- [ ] tests are implemented asserting logic is correct

```ts
interface Tag {
  name: string
  value: string
}

interface Cron {
  interval: string // '2-minutes'
  tags: Tag
}

function serializeCron (cron: Cron) => Tag[]
```

Happy path
```
GIVEN I am a developer using aoconnect
AND I have imported serializeCron
WHEN I pass valid interval and tags
THEN the parsed array of tags should be returned
AND the order of the tags should be preserved
```

Sad Path (invalid interval)
```
GIVEN I am a developer using aoconnect
AND I have imported serializeCron
WHEN I pass an invalid interval
THEN an error should be thrown indicating that my interval is invalid
```

Sad Path (invalid tags)
```
GIVEN I am a developer using aoconnect
AND I have imported serializeCron
WHEN I pass an invalid tag (both name and value must be a string)
THEN an error should be thrown indicating that tags are invalid
```

## Developer Notes

Will need to write the parse for the CU anyway. Would be a handy utility for the `aoconnect` to have to use along with `spawn`

For example, convert the JSON for a tags like:

```js
{
  interval: '10-minutes'
  tags: [
    { name: 'Foo', value: 'foo' },
    { name: 'Bar, value: 'bar' }
  ]
}
```

into tags that can be appended to the new process' tags:

```js
[
  { name: 'Cron-Interval', value: '10-minutes' },
  { name: 'Cron-Tag-Foo', value: 'foo' }
  { name: 'Cron-Tag-Bar', value: 'bar' }
]
```

jfrain99:
Closes #206 

Creates util function serializeCron which takes in a Cron Object and returns AO DataProtocol compliant tags.

### DIFF
--- a/connect/src/index.js
+++ b/connect/src/index.js
@@ -1,7 +1,6 @@
-import { connect } from './index.common.js'
+import { connect, serializeCron } from './index.common.js'
 
 import { WalletClient } from './client/node/index.js'
-
 const GATEWAY_URL = process.env.GATEWAY_URL || undefined
 const MU_URL = process.env.MU_URL || undefined
 const CU_URL = process.env.CU_URL || undefined
@@ -11,7 +10,7 @@ const { result, results, message, spawn, monitor, unmonitor, dryrun, assign } = 
 
 export { result, results, message, spawn, monitor, unmonitor, dryrun, assign }
 export { connect }
-
+export { serializeCron }
 /**
  * A function that builds a signer using a wallet jwk interface
  * commonly used in node-based dApps
--- a/connect/README.md
+++ b/connect/README.md
@@ -25,6 +25,7 @@ This module will run in a browser or server environment.
     - [`assign`](#assign)
     - [Environment Variables](#environment-variables)
     - [`createDataItemSigner`](#createdataitemsigner)
+    - [`serializeCron`](#serializecron)
 - [Debug Logging](#debug-logging)
 - [Testing](#testing)
 - [Project Structure](#project-structure)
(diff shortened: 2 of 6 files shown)


connect/src/index.js
```diff
@@ -1,6 +1,7 @@
 import { connect } from './index.common.js'
 
 import { WalletClient } from './client/node/index.js'
+import { serializeCron } from './lib/utils.js'
```
TillaTheHun0, line 4: `utils` is meant to be internal to the module, utilities that are used in the implementation, not so much _exported_ utilities. Otherwise, the entire sdk would be in `utils`.

Instead, there should be a `lib/serializeCron` and then place the impl in `index.js` there, then export in the top-lvl `index.js` from there.

connect/src/lib/utils.js
```diff
@@ -80,6 +80,40 @@ export function trimSlash (str = '') {
...
+ * @returns {{ name: string, value: string }[]}
+ */
+export function serializeCron (cron) {
+  function parseInterval (interval = '') {
+    if (typeof interval !== 'string') throw new Error('Encountered Error serializing cron: invalid interval')
+    const [value, unit] = interval.split('-').map(s => s.trim())
+    if (!value || !unit) throw new Error('Encountered Error serializing cron: invalid interval')
+    if (!parseInt(value) || parseInt(value) < 0) throw new Error('Encountered Error serializing cron: invalid interval value')
+
+    const regex = /(milliseconds|seconds|minutes|hours|days|months|years|blocks)/
```
TillaTheHun0, line 98: These can also be singular ie `1-hour` or `1-block`

connect/src/lib/utils.test.js
```diff
@@ -26,6 +26,98 @@ describe('utils', () => {
...
+            ]
+          }
+          try {
+            serializeCron(cronWrongInterval)
+          } catch (e) {
+            errorCaught = true
+            assert.equal(e.message, 'Encountered Error serializing cron: invalid interval')
+          }
+          assert.ok(errorCaught)
+        })
```
TillaTheHun0, line 47: Could you add spaces in between the tests? Just makes them a bit easier to read.

TillaTheHun0 (commented):
This is looking great, and looking close! A couple points of feedback, please see my comments.

TillaTheHun0 (approved):
LGTM

PR #745, opened 2024-05-29
perf(su): alter message table index for faster reads

jrdn91:
Adds migrations to remove the previous index for process_id and adds a new index combining process_id and timestamp. This speeds up queries to retrieve messages significantly.


VinceJuliano (approved):
LGTM!

jrdn91:
🚀

PR #750, opened 2024-05-29
feat(cu): add gas used as checkpoint condition #744

### LINKED ISSUE #744: feat(cu): add gas used as checkpoint condition
## Background

The CU generates checkpoints as a result of multiple events:
- if an eval stream length surpasses a threshold and on a certain time interval (both set in config).
- On a time-based interval, iff the `PROCESS_MEMORY_CACHE_CHECKPOINT_INTERVAL` is set
- On shut-down
- When a process is evicted from the LRU In-Memory cache

Each message evaluated by a process produces a `GasUsed` field.

## Problem 

All of these checkpoint mechanisms particularly the timespan and message-based events, provide at best, a fuzzy heuristic for a good point to checkpoint the process.

This results in Checkpoints being unnecessarily created, or evaluation stream still be long as a result of a process not being checkpointed enough.

We need a better indicator of when a process ought to be checkpointed to keep evaluation streams short, and subsequently request times low.

## Solution

We want checkpoints to be generated in a more useful way to minimize the amount of redundant evals needed. To do this, we will checkpoint based on gas used. Every x amount of gas used, the CU will create a checkpoint.

## Developer Notes

Makes sense to just keep a running tally of gas used on a process, and once it is greater than the threshold, create checkpoint and set gas used to 0.

jfrain99:
Closes #744 

Migrates the checkpoint threshold from a evalCount based system to a gas system. Now, if the process increases over the configurable gas threshold, a checkpoint will be saved and gas will reset to 0. Added tests to support.


servers/cu/src/domain/client/ao-process.js
```diff
@@ -1283,26 +1284,30 @@ export function saveLatestProcessMemoryWith ({ cache, logger, saveCheckpoint, EA
...
-      cron
+      cron,
+      gasUsed: incrementedGasUsed < EAGER_CHECKPOINT_ACCUMULATED_GAS_THRESHOLD ? incrementedGasUsed : 0
     }
     // cache.set(processId, { Memory: zipped, evaluation })
     cache.set(processId, { Memory, evaluation })
 
-    if (!evalCount || !EAGER_CHECKPOINT_THRESHOLD || evalCount < EAGER_CHECKPOINT_THRESHOLD) return
+    /**
+     *  Deprecated - we are no longer creating checkpoints based on eval counts. Rather, we use a gas-based checkpoint system
```
TillaTheHun0, line 1294: Could you use [JSDoc @deprecated](https://jsdoc.app/tags-deprecated)

servers/cu/src/config.js
```diff
@@ -104,6 +104,7 @@ const CONFIG_ENVS = {
     PROCESS_CHECKPOINT_CREATION_THROTTLE: process.env.PROCESS_CHECKPOINT_CREATION_THROTTLE || ms('30m'),
     DISABLE_PROCESS_CHECKPOINT_CREATION: process.env.DISABLE_PROCESS_CHECKPOINT_CREATION !== 'false',
     EAGER_CHECKPOINT_THRESHOLD: process.env.EAGER_CHECKPOINT_THRESHOLD || 100,
+    EAGER_CHECKPOINT_ACCUMULATED_GAS_THRESHOLD: process.env.EAGER_CHECKPOINT_ACCUMULATED_GAS_THRESHOLD || 300_000_000_000_000, // Amount of gas for 2 hours of continuous compute
```
TillaTheHun0, line 107: I'd like to see the reasoning behind _why_ this is ~2 hours of compute. Walk through the process used, for visibility and posterity. A good comments ought to describe the why

servers/cu/src/config.js
```diff
@@ -119,7 +120,7 @@ const CONFIG_ENVS = {
     PROCESS_MEMORY_CACHE_TTL: process.env.PROCESS_MEMORY_CACHE_TTL || ms('24h'),
     PROCESS_MEMORY_CACHE_DRAIN_TO_FILE_THRESHOLD: process.env.PROCESS_MEMORY_CACHE_DRAIN_TO_FILE_THRESHOLD || ms('5m'),
     PROCESS_MEMORY_CACHE_FILE_DIR: process.env.PROCESS_MEMORY_CACHE_FILE_DIR || tmpdir(),
-    PROCESS_MEMORY_CACHE_CHECKPOINT_INTERVAL: process.env.PROCESS_MEMORY_CACHE_CHECKPOINT_INTERVAL || ms('8h'),
+    PROCESS_MEMORY_CACHE_CHECKPOINT_INTERVAL: process.env.PROCESS_MEMORY_CACHE_CHECKPOINT_INTERVAL || 0,
```
TillaTheHun0, line 123: Did you update the README that describes the default value for this config?

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1267,6 +1267,7 @@ export function saveLatestProcessMemoryWith ({ cache, logger, saveCheckpoint, EA
     // const zipped = await gzipP(Memory, { level: zlibConstants.Z_BEST_SPEED })
     // stopTimer()
 
+    const incrementedGasUsed = (cached?.evaluation?.gasUsed ?? BigInt(0)) + (gasUsed ?? BigInt(0))
```
TillaTheHun0, line 1270: This is a little hard to read. Perhaps you could use a `pipe` with `ramda` utilities, such that it's more declarative? Just a suggestion.

servers/cu/src/domain/lib/evaluate.js
```diff
@@ -83,6 +83,7 @@ export function evaluateWith (env) {
 
   const saveLatestProcessMemory = saveLatestProcessMemorySchema.implement(env.saveLatestProcessMemory)
 
+  let totalGasUsed = BigInt(0)
```
TillaTheHun0, line 86: As it is, `totalGasUsed` would continue to accumulate across **_all evaluation streams, across all processes_**, eventually causing every eval stream to produce a checkpoint. Probably not what we want.

Instead this variable need to be moved into the function below, so that it is scoped to the single evaluation stream, then garbage collected. Perhaps within the function starting on ln.90

TillaTheHun0 (changes requested):

servers/cu/src/domain/lib/evaluate.js
jfrain99: 🤦

TillaTheHun0 (approved):
LGTM. Well done!

PR #765, opened 2024-06-04
fix(mu): throw error when .monitor same process twice

### LINKED ISSUE #752: fix(mu): it is possible again in the MU to run .monitor twice on the same process
image

jfrain99:
Closes #752 

Throws an error if you try to .monitor a cron process that is already being monitored.

### DIFF
--- a/servers/mu/src/domain/clients/cron.js
+++ b/servers/mu/src/domain/clients/cron.js
@@ -58,6 +58,9 @@ function initCronProcsWith ({ PROC_FILE_PATH, startMonitoredProcess }) {
 
 function startMonitoredProcessWith ({ logger, CRON_CURSOR_DIR, CU_URL, fetchCron, crank, PROC_FILE_PATH }) {
   return async ({ processId }) => {
+    if (cronsRunning[processId]) {
+      throw new Error('Process already being monitored')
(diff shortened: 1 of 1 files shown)


servers/mu/src/domain/clients/cron.js
```diff
@@ -58,6 +58,9 @@ function initCronProcsWith ({ PROC_FILE_PATH, startMonitoredProcess }) {
 
 function startMonitoredProcessWith ({ logger, CRON_CURSOR_DIR, CU_URL, fetchCron, crank, PROC_FILE_PATH }) {
   return async ({ processId }) => {
+    if (cronsRunning[processId]) {
+      throw new Error('Process already being monitored')
+    }
```
TillaTheHun0, line 63: How does this bubble up to client? Does the client receive a `4xx` status code? Also, is there an accompanying test for this? @VinceJuliano what ought to be the expected behavior here?

Perhaps the client should receive a `208`? Or perhaps more pessimistically a `409`?

VinceJuliano (approved):
Looks good to me, regarding Tyler's comment, I think we need to look at how errors are bubbling across the board but this is ok for now on the monitor

PR #766, opened 2024-06-04
feat(cu): allow checkpoint onwer to be configurable #758

### LINKED ISSUE #758: feat(cu): allow Checkpoint owner to be configurable
## Background

When the CU queries for Checkpoints it may use to hot start a process, it will only query for checkpoints created by itself. That is to say, Checkpoints created by it's wallet.

## Problem

If the CU would like to use a Checkpoint created by another owner, it is not possible

## Solution

Allow the list of owners the CU trusts for checkpoints to be configurable ie. an env var like `PROCESS_CHECKPOINT_TRUSTED_OWNERS`. This should use the `commaDelimitedArraySchema`.

If set, then the CU will query for checkpoints owned by these owners. If not set, then the CU will query checkpoints owned by itself.

> If a CU would like to trust itself and others, then it should simply pass it's own wallet along with any other wallets in the env var.

## Developer Considerations

`findLatestProcessMemory` will need to be updated to possibly use the env var. It's GQL query will need to be updated to accept `owners: [String!]!` (instead of only `owner: String!`). Then finally, the `variables` passed will need include `owners` not `owner`

jfrain99:
Closes #758 

If the PROCESS_CHECKPOINT_TRUSTED_OWNERS is set, use it to search Arweave. Otherwise, just use owner. Added test to ensure working.


servers/cu/src/domain/client/ao-process.js
```diff
@@ -732,12 +733,15 @@ export function findLatestProcessMemoryWith ({
...
-      }))
+      .chain((owner) => {
+        const owners = ifElse(isEmpty, always([owner]), always(PROCESS_CHECKPOINT_TRUSTED_OWNERS))(PROCESS_CHECKPOINT_TRUSTED_OWNERS)
+        return queryCheckpoints({
+          query: GET_AO_PROCESS_CHECKPOINTS,
+          variables: { owners, processId, limit: 50 },
+          processId,
+          before: LATEST
+        })
+      })
```
TillaTheHun0, line 744: `address()` only needs to be invoked if `PROCESS_CHECKPOINT_TRUSTED_OWNERS` is not defined. See if you can refactor to to take that into account. You'll need to use `of` and/or `Resolved`

TillaTheHun0 (commented):
Looks really close! I've made a suggestion to cut down on an invocation, please see comments.

jfrain99:
Resolved!

TillaTheHun0 (approved):
LGTM :shipit:

PR #771, opened 2024-06-05
perf(mu): cache potential CU redirect when fetching results #558

### LINKED ISSUE #558: perf(mu): cache potential CU redirect when fetching results
## Background

~We will soon be making~ We have made CU routers in testnet redirect to the underlying hosts in lieu of rev-proxying, to gain more visibility, and also align more closely with a decentralized network.

## Problem

The MUs requesting results from CUs will also follow these redirects received from the router. This extra redirect adds an extra "hop" to the MU -> CU comms. We could probably do something to mitigate this extra round trip.

## Solution

The MU implementation should be updated to cache these redirects received from the configured CU, if they occur, to eliminate a roundtrip when fetching results from a CU. Then on subsequent requests, the cache can be checked by `processId` and go directly to it's CU on a HIT.

jrdn91:
Closes #558 

Adds a new library utility to create a fetch function that is wrapped with a handler for redirects which writes to a passed in LRU cache so that subsequent requests will have their host set explicitly to the previously redirected url for the given process id

This is intended for use with the CU since URLs to be fetched include a processId in their query parameters and this is used as the cache key.

### DIFF
--- a/servers/mu/src/domain/clients/in-memory.js
+++ b/servers/mu/src/domain/clients/in-memory.js
@@ -3,12 +3,13 @@ import { LRUCache } from 'lru-cache'
 /**
  * @type {LRUCache}
  */
-export function createLruCache ({ size }) {
+export function createLruCache ({ size, ttl }) {
   const cache = new LRUCache({
     /**
      * number of entries
      */
     max: size,
+    ttl,
     /**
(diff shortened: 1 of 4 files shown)


VinceJuliano (approved):
LGTM

twilson63:
🚀

PR #776, opened 2024-06-06
refactor(cu): store file checkpoint data in sqlite instead of fs

### LINKED ISSUE #755: refactor(cu): store "file" checkpoint metadata in SQLite, and query it instead of glob on filesystem
## Background

When a CU creates a Checkpoint on Arweave, it will cache the location of that checkpoint in a "file checkpoint". This file doesn't contain the actual memory, but instead metadata concerning the checkpoint, and ultimately a pointer to the transaction on arweave. Historically, this has been used to reduce dependency on querying a gateway in order to find checkpoints, a bottleneck and potential point of failure ie. unstable gateway or delayed gateway indexing.

The files are named in a particular way to contain all the metadata needed to determine whether a file checkpoint is usuable.

The CU also has a SQLite Database.

## Problem

This glob can be expensive, and is also a bit esoteric. It also means that if the filesystem is swapped out, that  "file checkpoint cache" is lost.

## Solution

We can instead store the file checkpoint metadata and associated tx id in SQLite. Without backup of the SQLite, the same "what if the filesystem is swapped out" issue arises, but there are ways to restore just the SQLite database across filesystems.

## Developer Notes

This would involve creating a new table ie. `checkpoints` in SQLite, by updating this [file](https://github.com/permaweb/ao/blob/5ad575c1ef663e9004d5eafa346fca7ed4871a9a/servers/cu/src/domain/client/sqlite.js#L98)

Then creating a `checkpoints` model and doc schema, and refactoring usages in `ao-process.js`.

jfrain99:
Closes #755 

Refactors second stage of checkpoint looking from globbing a filesystem to a sqlite lookup. Creates record where previously created file. Queries records where previously queried files. Tests updated to reflect.


servers/cu/src/domain/client/ao-process.js
```diff
@@ -414,6 +417,97 @@ export function writeProcessMemoryFileWith ({ DIR, writeFile }) {
...
+  function createQuery ({ processId, before }) {
+    const timestamp = before === 'LATEST' ? new Date().getTime() : before.timestamp
+    return {
+      sql: `
+        SELECT *
+        FROM ${CHECKPOINTS_TABLE}
+        WHERE
+          processId = ? AND timestamp < ?
+        ORDER BY timestamp DESC
+        LIMIT 5;
```
TillaTheHun0, line 436: I'd place a comment explaining the rationale behind this query. Why `LIMIT 5`? Why both `processId` AND `timestamp`? How do this relate to the `latestCheckpointBefore` subsequent step.

servers/cu/src/domain/client/sqlite.js
```diff
@@ -65,6 +66,18 @@ const createMessages = async (db) => db.prepare(
...
+    id TEXT PRIMARY KEY,
+    processId TEXT,
+    timestamp INTEGER,
+    ordinate TEXT,
+    cron TEXT,
+    memory TEXT,
+    evaluation TEXT
+  ) WITHOUT ROWID;`
+).run()
+
```
TillaTheHun0, line 80: I wonder if we should also create an index on `processId` and `timestamp`?

servers/cu/src/domain/client/ao-process.js
```diff
@@ -414,6 +417,104 @@ export function writeProcessMemoryFileWith ({ DIR, writeFile }) {
...
+        evaluation.ordinate,
+        evaluation.cron,
+        JSON.stringify(Memory),
+        JSON.stringify(evaluation)
+      ]
+    }
+  }
+
+  return ({ Memory, evaluation }) => {
+    console.log(1, { Memory, evaluation })
```
TillaTheHun0, line 505: Stray `console.log`

servers/cu/src/domain/client/sqlite.js
```diff
@@ -78,6 +91,12 @@ const createMessagesIndexes = async (db) => db.prepare(
   `
 ).run()
 
+const createCheckpointIndexes = async (db) => db.prepare(
```
TillaTheHun0, line 94: `createCheckpointsIndexes` -- note the plural. Ought to match the table name.

TillaTheHun0 (approved):
LGTM. Awesome work! :shipit:

PR #779, opened 2024-06-08
fix(cu): allow checkpoint onwer to be configurable #758

7i7o:
This aims to fix a copy-paste error and adds a logger line to inform the trusted checkpoint wallets


jfrain99:
Thanks a ton!

PR #788, opened 2024-06-11
Jrdn91/process memory cache ttl

### LINKED ISSUE #666: `PROCESS_MEMORY_CACHE_TTL` cannot be large than `int32`
## Background

The CU LRU In-Memory cache TTL can be configured using the `PROCESS_MEMORY_CACHE_TTL`. Underneath the hood, a cached record's ttl is implemented using a simple `setTimeout` 

## Problem

NodeJS' `setTimeout` is implemented as a subroutine that uses a signed 32 bit integer to store the timeout time, in milliseconds. This means the maximum value for `PROCESS_MEMORY_CACHE_TTL` is 2^31 - 1 or ~25 days.

Setting `PROCESS_MEMORY_CACHE_TTL` to a higher value will cause Node to emit a warning and then implicitly overflow the integer. Subsequently, the `setTimeout` will be given a negative value and immediately be invoked. Basically disabling the LRU In-Mmeory cache.

## Solution

We'd like to allow larger values for `PROCESS_MEMORY_CACHE_TTL`. Can probably use something like [long-timeout](https://github.com/tellnes/long-timeout) which abstracts recursively calling `setTimeout` to allow for setting longer timeouts.

## Developer Notes

Code to modified is [here](https://github.com/permaweb/ao/blob/e30b00b5dd2d4357380064ce147f44fd6bcbc467/servers/cu/src/domain/client/ao-process.js#L43)

jrdn91:
closes #666 👿 

We are installing a new library `long-timeout` which can handle times longer than the max safe integer.
I have added a comment block to the main function to explain why this update is being made to force the code to pass in `setTimeout` as as dependency.

### DIFF
--- a/servers/cu/src/domain/client/ao-process.js
+++ b/servers/cu/src/domain/client/ao-process.js
@@ -47,7 +47,14 @@ export const LATEST = 'LATEST'
  * @prop {string} [cron]
  */
 let processMemoryCache
-export async function createProcessMemoryCache ({ MAX_SIZE, TTL, DRAIN_TO_FILE_THRESHOLD, gauge, onEviction, writeProcessMemoryFile }) {
+
+/**
+ * setTimeout is passed in as a dependency here because in a few instances such as TTL timeout, a value greater than the safe integer limit is used
+ * which causes setTimeout to overflow and return immediately. Allowing the caller to pass in their own setTimeout function allows them to handle this
+ * via passing in a new implementation such as using the 'long-timeout' library
+ * This resolves issue #666
+ */
+export async function createProcessMemoryCache ({ MAX_SIZE, TTL, DRAIN_TO_FILE_THRESHOLD, gauge, onEviction, writeProcessMemoryFile, setTimeout, clearTimeout }) {
   if (processMemoryCache) return processMemoryCache
 
   const clearTimerWith = (map) => (key) => {
--- a/servers/cu/src/domain/client/ao-process.test.js
+++ b/servers/cu/src/domain/client/ao-process.test.js
@@ -6,8 +6,9 @@ import { promisify } from 'node:util'
 
 import { createLogger } from '../logger.js'
 import { findLatestProcessMemorySchema, findProcessSchema, saveLatestProcessMemorySchema, saveProcessSchema } from '../dal.js'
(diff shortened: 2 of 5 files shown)


servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1287,5 +1288,42 @@ describe('ao-process', () => {
...
+        Memory: Buffer.from('hello world')
+      })
+      // wait to see that this implementation does not overflow
+      await new Promise((resolve) => setTimeout(resolve, 50))
+
+      assert.ok(setTimeoutCalled)
+      assert.ok(!ttlRan)
+
+      timeoutRef.unref()
+      lt.clearTimeout(timeoutRef)
```
TillaTheHun0, line 1324: If you inject `clearTimeout`, then I think you can call just `lt.clearTimeout` here.

TillaTheHun0 (commented):
Nice! A couple things, please see my comments.

servers/cu/src/domain/client/ao-process.js
```diff
@@ -47,7 +47,14 @@ export const LATEST = 'LATEST'
  * @prop {string} [cron]
  */
 let processMemoryCache
-export async function createProcessMemoryCache ({ MAX_SIZE, TTL, DRAIN_TO_FILE_THRESHOLD, gauge, onEviction, writeProcessMemoryFile }) {
+
+/**
+ * setTimeout is passed in as a dependency here because in a few instances such as TTL timeout, a value greater than the safe integer limit is used
+ * which causes setTimeout to overflow and return immediately. Allowing the caller to pass in their own setTimeout function allows them to handle this
+ * via passing in a new implementation such as using the 'long-timeout' library
+ * This resolves issue #6666
```
TillaTheHun0, line 55: `#666` instead of `#6666`

servers/cu/src/domain/index.js
```diff
@@ -185,7 +186,9 @@ export const createApis = async (ctx) => {
             value.evaluation,
             err
           )
-        })
+        }),
+    setTimeout: (...args) => lt.setTimeout(...args),
+    clearTimeout: lt.clearTimeout
```
TillaTheHun0, line 191: Same deal as above -- eta conversion may be needed, in case the dependency references `this`

TillaTheHun0 (approved):
LGTM :shipit:

Issue #802, opened 2024-06-18
perf(mu): Implement Application Http Metrics


jrdn91:
```
GIVEN I am a MU operator
WHEN I send a GET to the /metrics endpoint
THEN I should receive OpenMetrics detailing the http request total count and quantiles for the following endpoints:
- /monitor/:processId
- /
```

> This most likely is best implemented as a Histogram and/or Summary

PR #803, opened 2024-06-18
perf(mu): add metrics middleware for http requests #802

jrdn91:
This copies the `withMetrics` middleware from the CU and implements it across both the monitor endpoint and the root endpoint

### DIFF
--- a/servers/mu/src/routes/middleware/withMetrics.js
+++ b/servers/mu/src/routes/middleware/withMetrics.js
@@ -1,0 +1,48 @@
+import { always } from 'ramda'
+
+import { histogramWith } from '../../domain/clients/metrics.js'
+
+const histogram = histogramWith()({
+  name: 'http_request_duration_seconds',
+  description: 'Request duration in seconds',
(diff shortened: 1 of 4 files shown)


servers/mu/src/routes/middleware/withMetrics.js
```diff
@@ -0,0 +1,47 @@
+import { always } from 'ramda'
+import { histogramWith } from '../../domain/clients/metrics'
```
TillaTheHun0, line 2: Since the MU is type `module` this import requires an extension ie `/metrics.js`. I doubt the MU will start without throwing an error.

Also, a small convention is to add a space between it and dependencies resolved from `node_modules`. Not required, just helps keeps thing orderly.

TillaTheHun0 (commented):
Looks good, but please see my comment.

TillaTheHun0 (approved):
LGTM

TillaTheHun0 (approved):
LGTM :shipit:

PR #807, opened 2024-06-19
fix(mu): set MAX_WORKERS to possitiveIntSchema

jrdn91:
the MAX_WORKERS being set to `z.number()` throws a Zod error when starting up the MU because .env values are all strings, this change which mirrors the setting in the CU allows Zod to attempt to coerce the string to a positive integer or return a valid error for values that are not strings or cannot be cast to a positive integer


TillaTheHun0 (approved):
LGTM

PR #817, opened 2024-06-24
fix(mu): correct isWallet util function

### LINKED ISSUE #816: fix(mu): the mu is incorrectly sending messages to arweave
There are reported messages that the CU is unable to find when it goes to eval 

https://www.ao.link/#/message/Zj9DahInYDMOz587h4OeAUnjN50wW5Om396Gz3wmQiQ
https://www.ao.link/#/message/halQ48Z6KtL-ZftKDUl10De9ClmbWTT3BFPx0vZ1hGA

These messages can be found on chain via the gateway but not on the su. 

https://su8.ao-testnet.xyz/halQ48Z6KtL-ZftKDUl10De9ClmbWTT3BFPx0vZ1hGA&process-id=BUhZLMwQ6yZHguLtJYA5lLUa9LQzLXMXRfaq9FVcPJc

We believe this means the MU is sending them to the gateway instead of the su. The function isWalletWith in the gateway
client is likely recording a process as a wallet when it should not.

jfrain99:
Closes #816 

When determining if an id is a wallet or a process, we hit the gateway. If the gateway returns a 200, it is a process. Otherwise (error), it is a wallet. However, these calls can fail sometimes for whatever reason - so add a backoff on failure with 3 retries. If all 3 fail, wallet. Otherwise, process.

Additionally, add TTL to cache so that erroneous results don't last forever.

### DIFF
--- a/servers/mu/src/domain/clients/gateway.js
+++ b/servers/mu/src/domain/clients/gateway.js
@@ -1,3 +1,5 @@
+import { backoff, okRes } from '../utils.js'
+
 function isWalletWith ({ fetch, GRAPHQL_URL, logger, setById, getById }) {
   return async (id) => {
     logger(`Checking if id is a wallet ${id}`)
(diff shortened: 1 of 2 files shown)


VinceJuliano (approved):
LGTM!

PR #820, opened 2024-06-25
fix(mu): empty procs.json error catch

### LINKED ISSUE #811: fix(mu): procs.json was somehow emptied out
A mu server was unable to boot up. Upon further inspection the procs.json that controls cron was empty, somehow it was emptied out.

It is possible that when all processes are unmonitored it gets emptied out. See if we can reproduce.

jfrain99:
Closes #811 

Unable to reproduce error where procs.json was emptied out. However, added try catch to parsing of file to catch error and replace with {}. Works with empty or corrupted procs file.


VinceJuliano (approved):
LGTM!

PR #821, opened 2024-06-25
fix(mu): better error message when unmonitoring untracked processes #796

### LINKED ISSUE #796: fix(mu): unmonitor call is crashing
Users are reporting an issue with a call to unmonitor via aoconnect.

See if this is reproducible and implement a fix in the mu.

image

jfrain99:
Closes #796 

Was able to recreate issue - error throws when unmonitoring an untracked process. The error message is fairly unclear, so added thrown error explaining issue. Generally, though, working as intending.


VinceJuliano (approved):
LGTM

PR #823, opened 2024-06-25
perf(mu): Implement outgoing http Metrics #826

### LINKED ISSUE #802: perf(mu): Implement Application Http Metrics

### LINKED ISSUE #826: perf(mu): Implement Outgoing Http Metrics
```
GIVEN I am a MU operator
THEN I should receive OpenMetrics detailing the number of requests being made to external sources and their respective response times. External sources include:

- Gateway
- CU
- SU
- etc.
```

jrdn91:
closes #826

This adds a new util for swallowing errors as well as a withTimingMetrics HOF that can be used to wrap and compose functions that can keep track of execution time from the start to the return of the function.

### DIFF
--- a/servers/mu/src/domain/index.js
+++ b/servers/mu/src/domain/index.js
@@ -34,6 +34,13 @@ export { createLogger }
 const FIVE_MINUTES_IN_MS = 1000 * 60 * 5
 const SIXTY_MINUTES_IN_MS = 1000 * 60 * 60
 const muRedirectCache = InMemoryClient.createLruCache({ size: 500, ttl: FIVE_MINUTES_IN_MS })
+const histogram = MetricsClient.histogramWith()({
+  name: 'outgoing_http_request_duration_seconds',
+  description: 'Outgoing request duration in seconds',
+  buckets: [0.5, 1, 3, 5, 10, 30],
+  labelNames: ['method', 'operation', 'errorType', 'status', 'statusGroup']
+})
+
 /**
  * A set of apis used by the express server
  * to send initial items and start the message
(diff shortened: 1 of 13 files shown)


servers/mu/src/domain/index.js
```diff
@@ -30,6 +30,13 @@ const { DataItem } = warpArBundles
 const createDataItem = (raw) => new DataItem(raw)
 export { createLogger }
 
+const histogram = MetricsClient.histogramWith()({
+  name: 'outgoing_http_request_duration_seconds',
+  description: 'Outgoing request duration in seconds',
+  buckets: [0.5, 1, 3, 5, 10, 30],
+  labelNames: ['method', 'route', 'status', 'statusGroup']
```
TillaTheHun0, line 37: Won't `labelNames` need to have `operation` in it? Otherwise, stopping the timer will produce an error, not recording a metric and failing silently due to `swallow`.

Perhaps `withTimerMetrics` needs something such that it will "fail fast" if the consumer wants to set a label that is not allowed for the specific timer. 🤔  Otherwise, the seemingly benign action of adding a label downstream will break the metrics reporting. We'd want high, early, signal for that.

TillaTheHun0: Maybe for now, we add `operation` but put a comment on `withTimerMetrics` explaining as much. That labels reported _must_ be a subset of the labels defined on the `timer`.

servers/mu/src/domain/utils.test.js
```diff
@@ -84,4 +84,35 @@ describe('utils', () => {
...
+      assert.equal(result, 'hello world')
+    })
+
+    test('should return undefined for a thrown error', () => {
+      const safeFunction = swallow(() => { throw new Error('This error should be swallowed') })
+
+      const result = safeFunction()
+
+      assert.strictEqual(result, undefined)
+    })
```
TillaTheHun0, line 116: Could we make the spacing between blocks consistent across these tests? Doesn't matter which, just as long as it's consistent.

servers/mu/src/routes/middleware/withMetrics.js
```diff
@@ -30,19 +23,33 @@ export const withMetrics = ({ labelsFrom = always({}), tracesFrom = always({}) }
...
+
+  return (handler) => withTimerMetrics({
+    timer: histogram,
+    startLabelsFrom: labelsFromReq,
+    stopLabelsFrom: labelsFromRes,
+    tracesFrom
+  })((req, res) => Promise.resolve()
+    .then(() => handler(req, res))
+    .then(() => res)
+  )
```
TillaTheHun0, line 54: SO cool!

TillaTheHun0 (approved):
LGTM. Nice work!

Issue #824, opened 2024-06-26
AoLoader cannot load binaries produced with the latest version of the dev-cli


ppedziwiatr:
MRE is here: https://github.com/warp-contracts/ao-redstone-oracle/tree/main?tab=readme-ov-file#ao-redstone-oracle

ppedziwiatr:
Building wity the latest cli version (`0.0.59`) and using the `wasm64-unknown-emscripten-draft_2024_02_15` fixes the issue.

PR #827, opened 2024-06-26
docs: latest version for the dev-cli

ppedziwiatr:
Probably "fixes" https://github.com/permaweb/ao/issues/824 ;) - the mapping in the deno.json does not contain the latest version (at the time of writing - `0.0.59`) of the CLI


TillaTheHun0 (approved):
LGTM, thank you!

IIRC, there were issues in the CI uploading to ArDrive, and so the uploads were being performed manually as a workaround, causing the discrepancy in the manifest file (which is updated as part of CI)

PR #831, opened 2024-06-27
fix(mu): message reliability  persist task queue

jfrain99:
Addresses #661 Part 2 - Queue Persistance.

Persist task queue in sqlite database across lifecycles.


VinceJuliano:
I think DB_URL should probably be DB_PATH because its not a URL in this case

VinceJuliano (commented):
I think DB_URL should probably be DB_PATH because its not a URL in this case

VinceJuliano (approved):

PR #835, opened 2024-06-28
fix(cu): transfer memory between worker and main thread

TillaTheHun0:
This PR makes it so that the CU will transfer the ArrayBuffer representing an ao process' memory between the main thread and worker thread. This in lieu of a structured clone being performed on the ArrayBuffer across threads -- the default behavior of comms via a MessageChannel.

This should reduce the evaluation times for all processes, but particularly processes with large memory foot prints. Additionally, this also circumvents an existing NodeJS bug that prevent structured cloning an ArrayBuffer >4GB.

### DIFF
--- a/servers/cu/src/domain/client/ao-module.js
+++ b/servers/cu/src/domain/client/ao-module.js
@@ -6,6 +6,9 @@ import { z } from 'zod'
 
 import { moduleSchema } from '../model.js'
 import { MODULES_TABLE } from './sqlite.js'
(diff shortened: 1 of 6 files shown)


TillaTheHun0:
There seems to be an issue with ArrayBuffers piling up in the heap. Need to figure out if there's a place where an ArrayBuffer out to be dereferenced and subsequently garbage collected.

TillaTheHun0:
See #837 for memory leak fix

PR #839, opened 2024-06-30
fix: In the event of having no authorities on the list, trust no-one.

samcamwilliams:
While the default AOS behavior is now to add an authority explicitly while starting a process, there is still never a reason to make processes trust everyone in absence of an authority.

For all functional purposes, a process may as well be 'dead' than open for all to access. At some point perhaps a warning/confirmation method should be implemented if the user removes the last authority? But this should be carefully thought through.

For now, this code pattern is already being accidentally copied into other implementations (see the recent cpp dev-cli library), so let's remove the potential for danger from the 'reference' implementation.

### DIFF
--- a/dev-cli/container/src/ao.lua
+++ b/dev-cli/container/src/ao.lua
@@ -175,9 +175,9 @@ function ao.assign(assignment)
     table.insert(ao.outbox.Assignments, assignment)
 end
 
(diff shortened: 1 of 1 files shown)


TillaTheHun0 (approved):
Agreed. LGTM

PR #840, opened 2024-07-01
Erlang-style improved message handling

samcamwilliams:
This PR adds a callback to the return object in the `spawn` method. It will let devs easily perform actions once they have the process ID of the spawned message. This is critical to virtually every flow where spawning a process on-the-fly would be useful.

Notes:
- This PR requires the improved AOS `Handlers` PR to be merged in order for it to work.
- We need to use the `Ref_` (or similar) value to track our process after it was spawned. `Ref_` is a very ugly name -- do we know if anything is actually using it? If not, is now a good time to make it suck less?
- I wanted to simply _add_ the `after` function to the already returned table, but at the moment I cannot make that work. I believe the problem is Lua's local scoping? Any ideas?


samcamwilliams:
Added some more functionality in order to introduce Erlang-style pattern matching semantics to AOS:

- `Handlers` library now supports `resolvers` with multiple patterns and clauses in the matches. These allow us to create something akin Erlang's `receive ... end` statements in AOS. It also allows us to add/remove the handlers for a module or app within a single 'group', as well as have 'one time' handler groups in which a single match triggering automatically removes all of the other handlers, too. This is very helpful for easily building systems that respond to success/failure etc in another process.
- After sending a message, the user now receives an `.onReply()` callback that they can utilize to continue execution after receiving a reply. This function can take a `resolver` with multiple patterns and functions, so that the client can conditionally match on the response from the other process.
- All messages received by handlers now have `.reply()` and `.forward()` functions which allow the process to easily return a response to the user without having to manually deal with the `X-Reference` and `Target` values. This mechanism also allows the developer to easily build pipelines that process a request through many processes, then eventually return to the user.

This functionality is best understood through examples:

```lua
-- This test uses the new `onReply` callback function which is given to the
-- caller every time they send a function. `onReply` uses the outgoing message's
-- `Reference` number in order to catch a reply from the recipient via a handler.
function test_client_1(target)
    Test = nil
    ao.send({
        Target = target,
        Test = "1"
    }).onReply(
        function(m)
            print("Got reply!")
            print(m.Test)
            Test = "passed"
        end
    )
end

-- The test's counterparty is a simple server that receives the client's test
-- and uses each message's new `.reply` function in order to send a response.
-- The reply function extracts the `Reference` from the message and returns it
-- as an `X-Reference`, such that it can be tracked and responded to.
function test_server()
    Test = nil
    Handlers.once(
        { Test = "1" },
        function(m)
            print("Got message! Replying...")
            m.reply({ Test = "1-reply", Status = "Success" })
            Test = "replied"
        end
    )
end

-- This test creates a reply handler with multiple matcher patterns.
-- Both patterns and functions are added under a single 'run-once' handler.
-- Under the hood, `onReply` is using the `Handlers` library's new 'resolver'
-- functionality. Resolvers let us specify any pattern to match, and an
-- associated function to run, in a very similar fashion to Erlang's 'receive'
-- statements.
function test_client_2(target)
    Test = nil
    ao.send({
        Target = target,
        Test = "1"
    }).onReply({
        [{ Status = "Failure" }] =
            function(m)
                Test = "failed"
            end,
        [{ Status = "Success" }] =
            function (m)
                print("Got reply on pattern-matched handler!")
                Test = "passed"
            end
        }
    )
end

-- In this test we send a message to another process, who in turn forwards it
-- to another server, who then replies to the original sender. The forwarding
-- process thus acts as a 'bridge'. This pattern can be useful for building
-- 'pipelines' in which work is done by many processes in sequence, with the
-- response returned by the final participant.
function test_client_3(forwarder, replyer)
    Test = nil
    ao.send({
        Target = forwarder,
        Test = "1"
    }).onReply(replyer,
        function(m)
            print("Got reply from the end-of-pipeline!")
            print(m.Test)
            Test = "passed"
        end
    )
end

-- The `forwarder` process uses the new `.forward` function in order to push
-- all of the tags and data from the initial process onwards, along with the
-- client's original `X-Reference`. This `X-Reference` allows the client to
-- pattern match on the incoming message and run the associated handler,
-- even though the message the message was not originally sent to the replyer.
function test_forwarder(replyer)
    Test = nil
    Handlers.once(
        { Test = "1" },
        function(m)
            print("Got message! Forwarding...")
            m.forward(replyer)
            Test = "forwarded"
        end
    )
end
```

allquantor:
That's pretty cool @samcamwilliams 
Any thoughts on adding a `timeout` to the reply pattern?

samcamwilliams:
Update: Added `Receive` and `.receive()`s, such that processes can pseudo-synchronously process new messages without adding complexity to their control flow.

The basic `Receive()` function takes a pattern specification (see prior improvements in this and the [associated AOS PR](https://github.com/permaweb/aos/pull/281)) and halts execution of the current function until the first matching message is found. For example:

```lua
normal_processing()
Send({ Target = Process, Action = "Do-Something" })
Msg = Receive({ From = Process, Action = "Did-Something" }) -- Execution pauses here until response is received
print("Got message" .. Msg.id) -- Only prints _after_ the matching message has been received
continue_processing()
```

This allows the developer to write Erlang-style programs that interact with other processes without generating [callback hell](https://www.youtube.com/watch?v=NOlOw03qBfw).

Under-the-hood, `Receive` uses Lua's coroutines to pause and then resume execution of the function for handling a message. Notably, this scheme deviates from Erlang's 'blocking' `receive ... end` pattern by allowing other handlers to run in the intervening period between a `Receive` starting and ending. Each 'incomplete' handler is essentially added to a queue of coroutines that are waiting to receive their next message before continuing.

This PR also also implements `.receive()` functions for spawning processes and sending messages. These functions wrap the underlying `Receive()` to wait for the response to the interaction -- receiving the details of the newly spawned process, or the _reply_ (see details [here](https://github.com/permaweb/ao/pull/840#issuecomment-2205087214)).

For example:

```lua
ID = Spawn(MyModule, { Purpose = "Test" }).receive()["AO-Spawn-Success"] -- Execution pauses here until response is received
Result = Send({ Target = ID, Action = "Do-Something" }).receive() -- Send a request and wait for the response
print("Result of computation:" .. Result.Data)
```

These commits complete the API I had planned for the new process/message handling API. I _think_ the only  potential 'breaking' change is that `Spawn` and `Send` now have return values (so that users can access their returned functions), but it is hard to imagine that anything actually depends upon `Send(...) == nil` etc.

You can spawn live AOS processes using this new API as follows:

```bash
aos test-process-name --module=PzEpU-g7JbKTStMc8ov63QUT-PMfIZPtLO9BHVyYhkY
```

Open issues:
- Testing! I have been playing with this functionality live on the testnet, but there are no unit tests yet.
- Message prints seem to be missing (sometimes?). This is likely a missing `return` somewhere in the handler/coroutine management code. Run `CompleteTest2` in this [gist](https://gist.github.com/samcamwilliams/4ebff7463b12ecdc4a876095a82f7e19) to see an example.
- Naming. `receive` is the name that Erlang uses for this functionality, but it reads a little weirdly for newbies. In particular `.after(callback)` and `.receive()` on `Spawn` does not necessarily intuitively imply 'run this function after the process has spawned' and 'wait for this process to spawn'.
- Should `Receive` possibly scan the `Inbox` backwards to see if it matches with any other messages instantly? If it did, this would allow the order of the receives to be arbitrary. For example, imagine you are spawning 1,000 processes and waiting for them to reply with a computation result. If you ran a `Receive` for each in a loop, if the order you received them in was wrong some receives would not match. If `Receive` scanned the inbox they would 'buffer' there until their slot in the loop.
- Should `Receive` also optionally work with the same arguments as `Handlers.once({ [pattern] = function })`? This would make its functionality closer to Erlang and allow for synchronous conditional message handling, followed by further computation. Perhaps a `Handlers.add(...).receive()` to enforce waiting until matching would work?
- Particularly if we do the above, should `Receive` also be a single operand operator? Then you could say `msg = receive { Action = "Something" }`. This would be extremely simple syntax. If it is possible, I think an Erlang-style `ProcessID ! Msg` would be a great short-hand for `Send({ Target = ProcessID, ... })`.

twilson63:
> Testing! I have been playing with this functionality live on the testnet, but there are no unit tests yet.

Yep, we should implement unit tests for all of this in the AOS repo

> Naming. receive is the name that Erlang uses for this functionality, but it reads a little weirdly for newbies. In particular .after(callback) and .receive() on Spawn does not necessarily intuitively imply 'run this function after the process has spawned' and 'wait for this process to spawn'.

It may be worth considering `.await()` since many devs understand this from web/dev perspective, but I don't think it is a big deal.  Do we envision these functions to be `chainable` - if so, maybe it is worth considering functional patterns like `.map()` or `.fold()` 

> What if I just want to send a message with no `receive` or `onReply`?

```
Send({Target = ID, Data = "Hello"})
```

It looks like it goes through but the feedback is an object 

```
{
   onReply = function: 0x28e4600,
   receive = function: 0x28e4820
}
``` 

vs

```
Message added to outbox.
```

Do you always have to call `.receive()`

samcamwilliams:
Nope -- if you don't call `.receive()` it works like it used to :)

I think the message should still be displayed, but you are hitting that 'non-printing' bug @twilson63 ? Now it just returns the callbacks _as well_ as printing. I am not sure that we really want to print in general? Otherwise, we are pushing people to use `ao.[send|spawn]` in their real apps, and something different on the console.

That said, I also like that that when you are using it in 'CLI mode' you get positive feedback -- not just a new line (or worse, `undefined`). I wonder if the returned table with very obvious function names in it is enough to give the user a sense of what has happened?

WDYT?

twilson63:
bumped CLI Version

PR #842, opened 2024-07-01
fix(mu): add error assertion for missing metrics dependency

jrdn91:
This PR adds an early error check for the timer not being passed in or passed incorrectly for the metrics timer implementation. This will allow for an early signal on server startup if there is a timer metrics use case somewhere that a timer did not get injected into correctly.


TillaTheHun0 (approved):
Excellent change. LGTM :shipit:

PR #849, opened 2024-07-02
Jfrain99/task queue retry on error

### LINKED ISSUE #661: fix(mu): MU Message Reliability

jfrain99:
Closes #661 Number 1.

Adds a staging system to process result in the worker task queue. Keep a stage value in the ctx, which is updated as the process works. On an error, the task will be added back to the task queue and retried after 1000 ms a maximum of 5 times. Upon retry, tasks will 'skip' stages that have already been completed.


VinceJuliano (approved):
LGTM this is what we were going for

PR #851, opened 2024-07-02
perf(mu): adds gauge metrics for worker queue #830

### LINKED ISSUE #830: perf(mu): Implement Push Queue Size Metrics #802
GIVEN I am a MU operator
THEN I should receive OpenMetrics detailing the number of Data Items in the Push Queue

jrdn91:
closes #830 

Add a `BroadcastChannel` so have a worker post to the parent process as items are enqueued and dequeued. A gauge is passed to a broadcast message handler function to handle metrics.

A separate function has been used to handle metrics so that in the future different metrics / different broadcasted messages could be sent from the worker and handled differently in the parent process.


VinceJuliano (approved):
LGTM

PR #863, opened 2024-07-05
feat(cu): derive address from owner

### LINKED ISSUE #844: feat(cu): wrap chaining of evaluation streams determination as a critical section
## Background

The CU does it's best to deduplicate work across evaluation streams (see #627 for a brief summary of the process and #753 for how dry-runs are made non-blocking)

One of the first steps is determining whether an eval stream ought to be "chained" to an existing one. **This check is asynchronous**

## Problem

In certain situations, for example when many dry-runs are received at near the same time or even when multiple `/result` calls are received near the same time, because the check on whether to chain or not is asynchronous, the result can be multiple _duplicate_  or at the very best _overlapping_ evaluation streams.

This can place strain on the SU, and use up valuable compute resources on the CU.

## Solution

The `chain` check for a given ao process should be considered a critical section; only one actor may enter it at a time. This will serialize the checks and make chaining more effective, an overall boon for the entire system.

## Developer Notes

This would involve wrapping `chainEvaluation` step in `readState` with a mutex. The CU already leverages an implementation called `AsyncLock`, which is not strictly a lock and more of a access queue, but the desired results can be achieved with it.

### LINKED ISSUE #862: feat(cu): allow `Owner` and potentially `From` to be idiomatic etherium address
## Background

Currently, the CU impl sets `Owner` on messages assignments and processes and `From` on messages and assignments to the `owner.address`. 

This works great for owners that arweave wallets.

## Problem

Networks like Etherium [derive the wallet address](https://ethereum.org/en/developers/docs/accounts/#account-creation) and so using `owner.address` is not so idiomatic when a non-arweave wallet is used to sign an ao DataItem.

## Solution

The CU should examine the public key, and use that to determine which "format" fields like `Owner` and `From` ought to be.

Starting out, this will just be implemented for Etherium signers, but can be expanded to support other networks as well.

TillaTheHun0:
Closes #862 

This PR makes it so the CU will now set fields derived from `owner` using a projection of the data, instead of straight mapping `owner.address`. This allows the CU to map `owner` to more idiomatic addresses, given the owner. For example, this PR implements setting fields to an idiomatic Ethereum address if it detects the owner is an Ethereum account

Subsequently, the CU now needs more information to determine the value of fields on `Process`, `Module`, and `Message`, and that information was not being cached on the cached records. So this PR also includes changes such that the mapping of `owner` is performed in business logic, instead of pre-cached, and the cached records will "self-migrate" such that they store the necessary data needed by the CU to perform its derivations.

### DIFF
--- a/servers/cu/src/domain/utils.js
+++ b/servers/cu/src/domain/utils.js
@@ -1,3 +1,5 @@
+import { LRUCache } from 'lru-cache'
+import createKeccakHash from 'keccak'
 import { Rejected, Resolved } from 'hyper-async'
 import {
   F, T, __, allPass, always, append, assoc, chain, concat, cond, defaultTo, equals,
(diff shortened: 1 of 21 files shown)


servers/cu/src/domain/utils.js
```diff
@@ -422,3 +424,73 @@ export const maybeParseInt = pipe(
...
+  }
+}
+
+/**
+ * From etherium docs: https://ethereum.org/en/developers/docs/accounts/#account-creation
+ *
+ * You get a public address by taking the last 20 bytes of
+ * the Keccak-256 hash of the public key and adding 0x to the beginning.
+ */
+function keyToEtheriumAddress (key) {
```
samcamwilliams, line 456: *Ethereum

servers/cu/src/domain/utils.js
```diff
@@ -422,3 +424,73 @@ export const maybeParseInt = pipe(
...
+ * @returns {string}
+ */
+export const addressFrom = ({ address, key }) => {
+  const cKey = `${address}-${key}`
+  if (cache.has(cKey)) return cache.get(cKey)
+
+  const pubKeyLength = Buffer.from(key, 'base64url').length
+  let _address
+
+  if (pubKeyLength === SIG_CONFIG.ETHEREUM.pubLength) _address = keyToEtheriumAddress(key)
```
zyjblockchain, line 484: Determining the signature type of a bundle item by the length of its public key is not a good way to do this.
With the ANS104 standard, the bundle item has a signature type field to mark the type of public key used for signing.
ANS104: https://github.com/joshbenaron/arweave-standards/blob/ans104/ans/ANS-104.md
So it is possible to determine the address type simply by signatur type

zyjblockchain: image
This situation would be impossible to judge

TillaTheHun0: You're right, and I agree, that using signature type would be ideal. 

The impediment is that the CU actually does not handle raw Data Items directly, but retrieves message/process metadata from either the SU scheduling the process, or an Arweave Gateway (in the case of assignments), both of which do not currently expose the signature type. So currently, the CU simply does not have the signature type handy to make that distinction. However, it does have the pub key.

Using pub key length isn't ideal, or foolproof when considering all signature types, but it does allow for distinguishing Arweave from Ethereum, which is the focus of this change. It of course doesn't work when other signature types with same key lengths are considered, but then the impl just behaves as it does now (see the TODO on the fall through case in `addressFrom`). In this sense, these changes are additive. 

Ideally, the SU and gateway schemas can be extended to expose signature type, at which point the CU can and ought to use that to make this distinction. Until then, this gets what we need.

zyjblockchain: If this is the case then indeed the fastest and easiest way to integrate ethereum addresses is to use public length.
Hopefully this pr will be merged soon. I proposed support for ethereum address in an issue two days ago, and I didn't realise you were already implementing it. Thank you very much.
If this feature is merged into ao, the ao cross-chain bridge will natively support ethereum addresses, and users will be able to use the metamask to sign and query transactions and balances on ao, which is great for the whole ao ecosystem!

zyjblockchain:
When will this pr be merged to the official environment for use?

TillaTheHun0:
@zyjblockchain we still have to verify this doesn't break anything in the wider ecosystem, but i've been pulled onto other issues. Sorry for the delay, we'll try to get it merged as soon as we're able.

zyjblockchain:
> @zyjblockchain we still have to verify this doesn't break anything in the wider ecosystem, but i've been pulled onto other issues. Sorry for the delay, we'll try to get it merged as soon as we're able.

Thank you very much, it's very important to bring more users in when this feature goes live!

zyjblockchain:
Hello Could you please let me know when this pull request can be merged into the production environment?

TillaTheHun0:
@zyjblockchain sorry for the delays. There is a lot going on in the ao ecosystem that we've been focusing on and supporting. Hoping to have this merged soon.

zyjblockchain:
> @zyjblockchain sorry for the delays. There is a lot going on in the ao ecosystem that we've been focusing on and supporting. Hoping to have this merged soon.

AOX has completed code development from ethereum and bnbchain across the chain to AO and is now in the testing phase. Now we are waiting for this pr merge, AO natively supports 0x addresses, and AOX can help ethereum users to enter AO ecosystem without any obstacles.

Issue #593, opened 2024-04-05
FEATURE[aoconnect]: type exports in build


atticusofsparta:
Would love to be able to pull in the types in aoconnect in my application. Helpfull in FP with typescript to be able to set those types on data providers and functions.

TillaTheHun0:
Thanks for the suggestion. I thought we had this. `aoconnect` does specify type files [in the manifest](https://github.com/permaweb/ao/blob/d859d9779805ed9ae94b057cd2d1bc7632b7923c/connect/package.json#L23), but I guess that doesn't get the types.

The project isn't written in TypeScript. We generate ts declarations using tsc + JSDocs, and build the JS using `esbuild`. I'm not sure how to enable this with our toolchain, but it must _somehow_ be possible.

PRs are welcome.

dtfiedler:
I can offer a suggestion @TillaTheHun0

TillaTheHun0:
> I can offer a suggestion @TillaTheHun0

Please do! I am all ears :)

jim-toth:
Hey, I think there's a simple fix!  In `package.json` change:
`  "types": "./dist/index.d.ts",`
to
`  "types": "./dist/index.d.cts"`

jim-toth:
After attempting to use said types it seems they need to be updated?  For example, docs for loader opts have `inputEncoding` whereas in the type interface for Options it only has `input`.  Should these be updated to reflect the docs?  The generated types are out of date perhaps?

TillaTheHun0:
@jim-toth this issue is to do with [`aoconnect`](https://github.com/permaweb/ao/tree/main/connect), which is [this package](https://www.npmjs.com/package/@permaweb/aoconnect) on npm.

It sounds like you're referring to [`ao-loader`](https://github.com/permaweb/ao/tree/main/loader), which is [this package](https://www.npmjs.com/package/@permaweb/ao-loader) on npm. With that said, you're probably right that the types for that package need to be updated. Thank you for the suggestion for `ao-loader`! But this issue should stay focused on `aoconnect`.

TillaTheHun0:
Created an issue for updating and correcting `ao-loader` types #873. Thanks again for pointing that out!

PR #870, opened 2024-07-09
Fix typescript exports for ao-loader

jim-toth:


TillaTheHun0:
Thanks for taking a look @jim-toth.

Maybe I am missing something, but I don't see how this fixes the issue. Building `aoconnect` uses `tsc` to build and emit the type declarations, as `*.d.ts` files, into `/dist`; there are no `.cts` files generated into the `dist` folder. My understanding is that, if anything, this would break the current types.

What #593 is concerned with is the ability to import the types directly, in a TS project.

jim-toth:
> Thanks for taking a look @jim-toth.
> 
> Maybe I am missing something, but I don't see how the issue is addressed with these changes. Building `aoconnect` uses `tsc` to build and emit the type declarations, as `*.d.ts` files, into `/dist`; there are no `.cts` files generated into the `dist` folder. My understanding is that, if anything, this would break the current types.
> 
> What #593 is concerned with is the ability to import the types directly, in a TS project.

Well, I wasn't making a comment on the setup or what the files should be - just making a note of what worked for me.  That said, the types generated from the JSDoc comments in the source are out of date, so even getting the types imported correctly doesn't help much when they don't match the implementation.  I could open a PR or modify my PR to fix the JSDoc comments and thus generated types if you could provide the current interfaces - as i'm not as familiar with the implemented code.

Anyway, the `.cts` solution was because if you look at the package from npm, it has only two files one being a `*.d.cts*` which is why typescript can't find the files when it was told to look for `*.d.ts`.

See here for latest npm package with the `index.d.cts` file: https://www.npmjs.com/package/@permaweb/ao-loader?activeTab=code

See here for `package.json` explicitly looking for `index.d.ts`: https://github.com/permaweb/ao/blob/main/loader/package.json#L12

TillaTheHun0:
Got it, okay I see. These changes are to do with `ao-loader` and not `aoconnect`. Yes, it does look like the `package.json` file references the wrong file for types, and to your point, those types are most likely out of date anyway (`ao-loader` is under a lot of flux!).

We will try to get those types updated on `ao-loader`, will create an issue for that. Thank you for pointing it out!

jim-toth:
> Got it, okay I see. These changes are to do with `ao-loader` and not `aoconnect`.

Yep, and when using `@permaweb/ao-loader` the npm package points to this repository...which looks like a monorepo...

Is there another repository dedicated to `ao-loader` ?  Then the package.json should be updated to reflect this, so others don't wind up in this repo?

jim-toth:
> @jim-toth yes it's a monorepo. `ao-loader` is [here](https://github.com/permaweb/ao/tree/main/loader)

Gotcha, sorry for the confusion!  We got there in the end 🤝

TillaTheHun0:
Types for `aoloader` were updated in #954. @jim-toth can this be closed?

TillaTheHun0:
Closing this for now, as I am dong some housekeeping. If I am misunderstanding, please let me and I can reopen. Thanks for bringing the types to our attention.

PR #872, opened 2024-07-11
Jfrain99/persist lru cache in fs 843

### LINKED ISSUE #843: feat(cu): persist LRU In-Memory cache process state across restarts
## Background

The CU utilizes an LRU In-Memory cache for keeping ao processes hot, storing each ao process' latest known Memory, and using that to resume WebAssembly Instances.

Part of this cache involves a mechanism for draining LRU processes to the file system, updating the cache entry with a reference to that memory, and freeing up space in the heap for more "active" processes. See #837 for details

## Problem

On CU restarts, the LRU In-Memory cache entries are lost, and thereby any references to process memory drained to the filesystem. The result is that CU crashes, or restarts that do not successfully checkpoint on Arweave, are not able to use this readily available resource, for lack of a way to retrieve and check it's relevancy -- the corresponding metadata for the memory ie. the evaluation none, block-height, epoch, timestamp, ordinate, cron, etc.

It would be great if we could somehow leverage this data across restarts, the LRU In-Memory "rehydrating" itself using the drained entries on the file system.

An ancillary benefit of this is that the CU could feasibly checkpoint to Arweave less often, perhaps _only_ when accumulated gas used hits a certain threshold, ergo cutting back cost for the CU operator to persist checkpoints to Arweave.

## Solution

Modify the LRU In-Memory Cache behavior to drain entire cache entries to the filesystem (memory and metadata) and a mechanism to "rehydrate" across restarts.

## Developer Notes

The implementation could end up being _very_ similar to the original "file checkpoints" in the CU, except instead of `Memory.id` in the JSON pointing to a Checkpoint on Arweave, it could be `Memory.file` pointing to a file on the filesystem, and that file that contains the ao process memory.

So the "file checkpoint" metadata could be globbed for, and if found, the referenced file read and altogether used as the checkpoint -- a true "file checkpoint".

I suspect this could be another step in the `findLatestProcessMemory` flow, after `memory` checkpoint, but before `record` checkpoint.

Finally, `saveCheckpoint` could be made to write Record Checkpoints, but then also File Checkpoints to disk.

> Since the actually memory will be written to disk, this could fill up local disk quickly with old File Checkpoints. So it would be great if there was something cleaning those up periodically.

jfrain99:
Closes #843 

Adds a sqlite table that acts as a directory for checkpoint files. ~Whenever a checkpoint is drained/saved to a file, the directory updates.~

When finding latest process memory, before searching a cached record that points to a checkpoint on Arweave aka (a "record" checkpoint), the CU will look for a file checkpoint and use that if one is found.

File checkpoints will be created anytime the CU is compelled to create checkpoints, regardless or weather or not a checkpoint is uploaded to Arweave.


servers/cu/src/domain/index.js
```diff
@@ -142,15 +142,16 @@ export const createApis = async (ctx) => {
...
-    writeCheckpointFile: AoProcessClient.writeCheckpointFileWith({
-      DIR: ctx.PROCESS_CHECKPOINT_FILE_DIRECTORY,
-      writeFile
-    }),
     writeCheckpointRecord: AoProcessClient.writeCheckpointRecordWith({
       db: sqlite
     }),
+    writeFileRecord: AoProcessClient.writeFileRecordWith({
+      db: sqlite
+    }),
```
TillaTheHun0, line 150: You can create this side effect once, then inject it into multiple places.

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1,8 +1,7 @@
 import { promisify } from 'node:util'
 import { gunzip, gzip, constants as zlibConstants } from 'node:zlib'
 import { Readable } from 'node:stream'
-import { basename, join } from 'node:path'
-
+import { join } from 'node:path'
```
TillaTheHun0, line 4: Let's put a space between the `node:*` deps and other deps.

servers/cu/src/domain/client/ao-process.js
```diff
@@ -286,6 +292,11 @@ const processDocSchema = z.object({
 
 function latestCheckpointBefore (destination) {
   return (curLatest, checkpoint) => {
+    /**
+     * What we will compare. In some cases, the timestamp is in the evaluation
+     * key of the checkpoint object. This handles those cases.
+     */
+    const evaluation = checkpoint.evaluation ?? checkpoint
```
TillaTheHun0, line 299: What about the other fields that are compared? `ordinate` and `cron`?

TillaTheHun0: Any way to for `latestCheckpointBefore`'s `reduce` to have a single api, instead of supporting two different input shapes? Any way to adapt the caller to achieve that? Think the [open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1390,7 +1511,7 @@ export function saveCheckpointWith ({
     if (!recentCheckpoints.has(processId)) return Rejected(args)
 
     logger('Checkpoint was recently created for process "%s", and so not creating another one.', processId)
-    return Resolved()
+    return Rejected(args)
```
TillaTheHun0, line 1514: Did you mean to invert this `Resolved` into a `Rejected`?

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1466,6 +1593,48 @@ export function saveCheckpointWith ({
               return { id: res.id, encoding: 'gzip' }
             }
           )
+          .bichain(
+            (ctx) => {
+              /**
+               * If file checkpoints are disabled, skip this step.
+               */
+              if (DISABLE_PROCESS_FILE_CHECKPOINT_CREATION) return Rejected(ctx)
+              console.log(100, { ctx })
```
TillaTheHun0, line 1602: Stray `console.log`

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1466,6 +1593,48 @@ export function saveCheckpointWith ({
...
+                encoding: 'gzip'
+              }
+              if (Memory) {
+                writeProcessMemoryFile({
+                  Memory,
+                  evaluation
+                })
+                  .then((path) => {
+                    writeFileRecord(evaluation, path)
+                  })
```
TillaTheHun0, line 1625: Should probably be `.then((path) => writeFileRecord(evaluation, path))`. Otherwise, the `Promise` is left dangling and won't trigger the `catch` and instead bubble.

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1466,6 +1593,48 @@ export function saveCheckpointWith ({
...
+                timestamp,
+                nonce,
+                ordinate,
+                epoch,
+                blockHeight,
+                cron,
+                encoding: 'gzip'
+              }
+              if (Memory) {
+                writeProcessMemoryFile({
```
TillaTheHun0, line 1619: This `Promise` is simply dangling. Is this intentional? If so, there should be a comment clearly explaining why

servers/cu/src/domain/client/ao-process.js
```diff
@@ -1466,6 +1593,48 @@ export function saveCheckpointWith ({
...
+                  })
+                  .catch((err) => {
+                    logger(
+                      'Error occurred when draining Memory for evaluation "%j" to a file. Skipping adding back to the cache...',
+                      evaluation,
+                      err
+                    )
+                  })
+              }
+              return Rejected(ctx)
```
TillaTheHun0, line 1634: This strikes me as strange to return a `Rejected` here. Maybe I am misunderstanding?

servers/cu/src/domain/client/ao-process.js
```diff
@@ -854,6 +914,58 @@ export function findLatestProcessMemoryWith ({
       })
   }
 
+  /**
+   * Check if there is a process checkpoint in the file system.
+   * First, we will search the sqlite directory for a path and ensure the evaluation is before our current eval.
+   * Then, we will check if the file exists and read the file.
+   * Then, we will decode the data.
```
TillaTheHun0, line 921: I don't see where any decoding is done in this function.

servers/cu/src/domain/client/ao-process.js
```diff
@@ -854,6 +914,58 @@ export function findLatestProcessMemoryWith ({
       })
   }
 
+  /**
+   * Check if there is a process checkpoint in the file system.
+   * First, we will search the sqlite directory for a path and ensure the evaluation is before our current eval.
```
TillaTheHun0, line 919: What is the `sqlite` directory mean?

servers/cu/src/domain/client/ao-process.js
```diff
@@ -854,6 +912,64 @@ export function findLatestProcessMemoryWith ({
...
+        /**
+         * No previously created checkpoint is cached in a file,
+         * so keep looking
+         */
+        if (!checkpoint) return Rejected(args)
+
+        /**
+         * We have found a previously created checkpoint, cached in a record, so
+         * we can skip querying the gateway for it, and instead download the
+         * checkpointed Memory directly, from arweave.
```
jfrain99, line 937: fix

Issue #683, opened 2024-05-07
build(su): su.yml that was added is crashing the testing CI


VinceJuliano:
## Background

A su.yml file was added in a previous PR that uses a docker image we dont have access to

## Problem

This is causing the SU testing CI to record a failure

## Solution

Fix or remove this .yml file

## Developer Notes

We will have to make the docker image publicly accessible or modify the .yml

dtfiedler:
Related fix - https://github.com/permaweb/ao/pull/879

PR #879, opened 2024-07-11
fix(su): update su github to publish to permaweb ghcr

dtfiedler:
This was a copy/pasta that is causing publishing images to fail.


dtfiedler:
fixes #683

TillaTheHun0 (approved):
LGTM :shipit:

PR #881, opened 2024-07-12
Write test cases for saveCheckpointWith

### LINKED ISSUE #804: test(cu): implement tests for `saveCheckpointWith`
## Background

`ao-process.js` in the CU contains side effects pertaining to ao processes. One of these is `saveCheckpointWith`, which contains the logic for querying, building, signing, and uploading ao Process `Checkpoints` to Arweave, as well as caching them as a record in the SQLite DB (see #776)

## Problem

This api currently lacks unit tests asserting it's correctness

## Solution

Write the tests :)

jrdn91:
closes #804 

Adds a number of test cases to test each branch as well as the "happy path"

Adjusts the internal `recentCheckpoints` to be a dependency that defaults to a `new Map()`

Add a conditional check to verify that the Data Item has a tags array

### DIFF
--- a/servers/cu/src/domain/client/ao-process.test.js
+++ b/servers/cu/src/domain/client/ao-process.test.js
@@ -6,7 +6,7 @@ import { promisify } from 'node:util'
 
 import { createLogger } from '../logger.js'
 import { findLatestProcessMemorySchema, findProcessSchema, saveLatestProcessMemorySchema, saveProcessSchema } from '../dal.js'
-import { LATEST, createProcessMemoryCache, findCheckpointFileBeforeWith, findLatestProcessMemoryWith, findProcessWith, saveLatestProcessMemoryWith, saveProcessWith } from './ao-process.js'
+import { LATEST, createProcessMemoryCache, findCheckpointFileBeforeWith, findLatestProcessMemoryWith, findProcessWith, saveCheckpointWith, saveLatestProcessMemoryWith, saveProcessWith } from './ao-process.js'
 import { Readable } from 'node:stream'
 import bytes from 'bytes'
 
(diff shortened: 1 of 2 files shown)


servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1396,5 +1396,107 @@ describe('ao-process', () => {
...
+          return Promise.resolve()
+        },
+        DISABLE_PROCESS_CHECKPOINT_CREATION: false
+      })
+      try {
+        const result = await saveCheckpoint({ File: 'file', moduleId: 'module-123', processId: 'process-123' })
+        assert.equal(result.id, 'tx-123', 'should return the id of the checkpoint')
+      } catch (e) {
+        console.log('error', e)
+      }
```
TillaTheHun0, line 1499: Catching an error here will cause the test to always pass, even if an error thrown. We should remove the `try/catch`

servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1396,5 +1396,107 @@ describe('ao-process', () => {
...
+        } catch (e) {
+          assert.equal(e, 'either File or Memory required')
+        }
+      })
+      test('should fail if file promise resolves with undefined', async () => {
+        const saveCheckpoint = saveCheckpointWith({
+          ...deps
+        })
+        try {
+          await saveCheckpoint({ File: Promise.resolve(undefined) })
```
TillaTheHun0, line 1471: this should have something like `.then(() => assert.fail('should not resolve if File resolves to undefined'))` or something like that.

servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1396,5 +1396,107 @@ describe('ao-process', () => {
...
+        queryGateway: () => Promise.resolve(),
+        queryCheckpointGateway: () => Promise.resolve(),
+        DISABLE_PROCESS_CHECKPOINT_CREATION: false
+      }
+      test('should fail if memory and file are both undefined', async () => {
+        const saveCheckpoint = saveCheckpointWith({
+          ...deps
+        })
+        try {
+          await saveCheckpoint({})
```
TillaTheHun0, line 1461: this should have something like .then(() => assert.fail('should not resolve if no Memory or File is provided')) or something like that.

servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1396,5 +1396,107 @@ describe('ao-process', () => {
...
+    test('should not save checkpoint if process was recently checkpointed', async () => {
+      const saveCheckpoint = saveCheckpointWith({
+        ...depsAll,
+        address: () => {
+          assert.fail('saving checkpoint process should not have been initiated')
+        },
+        DISABLE_PROCESS_CHECKPOINT_CREATION: false,
+        recentCheckpoints: new Map([['process-123', true]])
+      })
+      await saveCheckpoint({ processId: 'process-123' })
```
TillaTheHun0, line 1425: This needs an assertion on the returned value

servers/cu/src/domain/client/ao-process.test.js
```diff
@@ -1396,5 +1396,107 @@ describe('ao-process', () => {
...
+
+    test('should not save checkpoint if checkpointing is disabled', async () => {
+      const saveCheckpoint = saveCheckpointWith({
+        ...depsAll,
+        address: () => {
+          assert.fail('saving checkpoint process should not have been initiated')
+        },
+        DISABLE_PROCESS_CHECKPOINT_CREATION: true
+      })
+      await saveCheckpoint({})
```
TillaTheHun0, line 1413: This should have an assertion on the returned value being `undefined` or whatever is returned.

TillaTheHun0 (commented):
Looking good, but please see my comments.

TillaTheHun0 (approved):
LGTM

PR #883, opened 2024-07-15
chore(cu): add backoff, circuit breaker to gatweay client for load block metadata

### LINKED ISSUE #368: chore(cu): wrap Gateway client in a CircuitBreaker/backoff, so to rate limit requests to the gateway and help prevent 429s
## Background

Related to #363 which implemented caching blocks, but not the circuit breaker.

The CU currently makes requests to the Gateway to fetch block metadata (for the purposes of generating cron messages) and process metadata (in order to initally locate a process on a SU).

Finally the CU will also call arweave directly in order to load assigned messages, as well as ao `Modules`

## Problem

The `ARWEAVE_URL` and `GRAPHQL_URL` can both start to return a status code of `429` to requests -- rate limiting requests from the CU

Currently the CU does not handle these responses gracefully

## Solution

We would like to add, at the least, a retry mechanism with an exponential backoff, to recover from 429s. Ideally, we can additionally employ a Circuit Breaker pattern to rate limit requests client side, as to avoid the 429s as much as possible.

## Developer Notes

- There is a utility in the CU, `backoff`, that implements a generic retry with exponential backoff
- [`opossum`](https://www.npmjs.com/package/opossum) is a good Circuit Breaker implementation in NodeJS we might could use.

jfrain99:
Closes #368 

Adds a backoff and a circuit breaker to load block metadata call. Adds tests to ensure they are working.

### DIFF
--- a/servers/cu/src/domain/client/ao-block.js
+++ b/servers/cu/src/domain/client/ao-block.js
@@ -2,9 +2,15 @@ import { fromPromise, of } from 'hyper-async'
 import { applySpec, last, map, path, pathSatisfies, pipe, pluck, prop, props, splitEvery } from 'ramda'
 import { z } from 'zod'
 import pMap from 'p-map'
-
 import { blockSchema } from '../model.js'
 import { BLOCKS_TABLE } from './sqlite.js'
+import { backoff, strFromFetchError } from '../utils.js'
+import CircuitBreaker from 'opossum'
+
+const okRes = (res) => {
+  if (res.ok) return res
+  throw res
+}
 
 const blockDocSchema = z.object({
   id: blockSchema.shape.height,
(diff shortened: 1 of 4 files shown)


jfrain99:
AOS error messages

servers/cu/src/domain/client/ao-block.js
```diff
@@ -146,16 +158,21 @@ export function loadBlocksMetaWith ({ fetch, GRAPHQL_URL, pageSize, logger }) {
...
+            () => fetch(GRAPHQL_URL, {
+              method: 'POST',
+              headers: { 'Content-Type': 'application/json' },
+              body: JSON.stringify({
+                query: GET_BLOCKS_QUERY,
+                variables
+              })
+            }).then(okRes).catch((_e) => {
+              throw new Error('Can not communicate with gateway to retrieve block metadata')
+            }),
```
TillaTheHun0, line 172: It would be nice to log the actual being caught, for posterity in the logs.

servers/cu/src/domain/client/ao-block.js
```diff
@@ -146,16 +158,21 @@ export function loadBlocksMetaWith ({ fetch, GRAPHQL_URL, pageSize, logger }) {
...
+            { maxRetries: 2, delay: 300, log: logger, name: `loadBlockMeta(${JSON.stringify({ GRAPHQL_URL, newMin, maxTimestamp })})` }
```
TillaTheHun0, line 173: Let's not print the GRAPHQL_URL here

TillaTheHun0 (commented):
Looks good, but just had a couple suggestions.

TillaTheHun0:
Also consider whether we should leverage `opossum`'s [other options](https://nodeshift.dev/opossum/#circuitbreaker). `allowWarmUp` and `volumeThreshold` stick out to me as potentially useful for use-case

TillaTheHun0 (approved):
LGTM

PR #886, opened 2024-07-17
perf(cu): remove wasm instances from cache on finish #845

### LINKED ISSUE #845: perf(cu): remove unused WebAssembly.Instance from worker thread cache
## Background

For each evaluation stream, the CU will instantiate a `WebAssembly.Instance`, caching it in an LRU In-Memory cache in the respective worker threads performing evaluations as part of the evaluation stream.

Each worker thread's cache is a simple LRU Cache with a max size configured by `WASM_INSTANCE_CACHE_MAX_SIZE`.

The cache is not "cleaned" up at the end of an evaluation stream, leaving the LRU eviction mechanism to remove unused `WebAssembly.Instance`s

## Problem

This means `WebAssembly.Instance`s from previously completed evaluation streams could "hang out" in each worker threads LRU In-Memory cache, indefinitely, bloating memory and taking up space in the heap. In certain situations, this can contribute to the CU running out of memory (a burst of traffic distributes work across many workers, bloating memory. Then traffic dies down, but the unused `WebAssembly.Instance`s remain. Gradually, the memory could be consumed by on instances, not being evicted because the worker threads are simply going unused, and so are not evicting their caches.

## Solution

Signal to the worker when the last evaluation for an eval stream is being received, so that it can "self cleanup", removing the `WebAssembly.Instance` from the worker thread's cache, allowing it to be garbage collected

This should free up significant memory after spikes in traffic.

## Developer Notes

`lib/evaluate.js` could be modified such that it knows when the final message is being emitted, and pass a `last` flag across to the worker. The worker could then examine this `last` argument and conditionally remove the WebAssembly.Instance from the cache, since it will no longer be used.

jfrain99:
Closes #845 

Sends a signal when the eval stream is finished. Delete WASM instance from cache when this flag is received.

### DIFF
--- a/servers/cu/src/domain/worker/evaluate.js
+++ b/servers/cu/src/domain/worker/evaluate.js
@@ -158,12 +158,20 @@ export function evaluateWith ({
    *
    * Finally, evaluates the message and returns the result of the evaluation.
    */
-  return ({ streamId, moduleId, wasmModule, moduleOptions, processId, noSave, name, deepHash, cron, ordinate, isAssignment, Memory, message, AoGlobal }) =>
+  return ({ streamId, moduleId, wasmModule, moduleOptions, processId, noSave, name, deepHash, cron, ordinate, isAssignment, Memory, message, AoGlobal, close }) => {
+    /**
+     * If we receive the close flag, it means we have reached the end of the eval stream.
+     * We will remove the WASM instance from the cache and skip loading the module.
+     */
+    if (close) {
+      wasmInstanceCache.delete(streamId)
+      return
+    }
     /**
      * Dynamically load the module, either from cache,
      * or from a file
      */
-    maybeCachedInstance({ streamId, moduleId, wasmModule, moduleOptions, name, processId, Memory, message, AoGlobal })
+    return maybeCachedInstance({ streamId, moduleId, wasmModule, moduleOptions, name, processId, Memory, message, AoGlobal })
       .bichain(loadInstance, Resolved)
       /**
        * Perform the evaluation
@@ -225,4 +233,5 @@ export function evaluateWith ({
           })
       )
       .toPromise()
+  }
 }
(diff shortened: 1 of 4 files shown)


servers/cu/src/domain/worker/evaluate.js
```diff
@@ -158,12 +158,16 @@ export function evaluateWith ({
...
     /**
      * Dynamically load the module, either from cache,
      * or from a file
      */
-    maybeCachedInstance({ streamId, moduleId, wasmModule, moduleOptions, name, processId, Memory, message, AoGlobal })
+    if (close) {
+      wasmInstanceCache.delete(streamId)
+      return
+    }
+    return maybeCachedInstance({ streamId, moduleId, wasmModule, moduleOptions, name, processId, Memory, message, AoGlobal })
```
TillaTheHun0, line 170: Please move this original comment down to it's accompanying line.

This `if` also needs it's own comment explaining what `close` is, when it will be passed, and the why behind this code.

servers/cu/src/domain/worker/evaluate.js
```diff
@@ -158,12 +158,20 @@ export function evaluateWith ({
...
-  return ({ streamId, moduleId, wasmModule, moduleOptions, processId, noSave, name, deepHash, cron, ordinate, isAssignment, Memory, message, AoGlobal }) =>
+  return ({ streamId, moduleId, wasmModule, moduleOptions, processId, noSave, name, deepHash, cron, ordinate, isAssignment, Memory, message, AoGlobal, close }) => {
+    /**
+     * If we receive the close flag, it means we have reached the end of the eval stream.
+     * We will remove the WASM instance from the cache and skip loading the module.
+     */
+    if (close) {
+      wasmInstanceCache.delete(streamId)
+      return
+    }
```
TillaTheHun0, line 169: Could there be a test for this branch in `worker/evaluate.test.js`? Can probably initialize the cache, inject it, call while passing `close` then assert the injected caches size was decremented by 1.

TillaTheHun0 (approved):
LGTM

PR #887, opened 2024-07-17
fix(scheduler-utils): trim trailing slash from returned URLs #246

### LINKED ISSUE #246: scheduler-utils: trim trailing slash from `Url` if it exists
so `Url` is a consistent shape that can be safely appended to

jrdn91:
closes #246 

adds a `trimTrailingSlash` function with a test to trim the trailing slash from the URL if it has one.

Wraps the URL returned from both `raw` and `locate` function return values.


TillaTheHun0 (approved):
LGTM :shipit:

PR #896, opened 2024-07-18
feat(mu): separate URL configuration

### LINKED ISSUE #894: fix(mu): separate out GRAPHQL_URL and GATEWAY_URL
The MU needs to use separate URLs for graphql and other gateway operations. Because in many cases the apis are on different domains.

TillaTheHun0:
Closes #894 

This PR separates `GRAPHQL_URL` into two separate configurations:

- `ARWEAVE_URL`: the host of the Arweave http server to be used by the MU
- `GRAPHQL_URL`: the host of the Arweave GraphQL server to be used by the MU

Before these two services were conflated, when they are actually separate services.

This functionality matches the behavior of the CU, by allowing a default `GATEWAY_URL` as a default configuration fallback that is used to default both values, if not specified.

Finally, `isWallet` was refactored to use the separate `ARWEAVE_URL`, in lieu of parsing the structure of `GRAPHQL_URL`, which was not only an incorrect conflation (as described above), but brittle in the the sense that it assumed a specific structure of the string

### DIFF
--- a/servers/mu/.env.example
+++ b/servers/mu/.env.example
@@ -1,6 +1,5 @@
 PORT=3004
 CU_URL="https://cu.ao-testnet.xyz"
-GRAPHQL_URL="https://arweave.net/graphql"
 NODE_CONFIG_ENV="development"
 DEBUG=*
 PATH_TO_WALLET="/home/wallet.json"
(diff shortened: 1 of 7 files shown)


VinceJuliano (approved):
LGTM

PR #901, opened 2024-07-19
test(mu): add tests for cron monitor functions

jrdn91:
This PR adds a refactor to the init monitor HOF so that it can be tested and adds some initial tests for the init function and the "kill" function.

TODO: tests still need to be added for `startMonitoredProcessWith` and it will probably need refactored to add dependency injection to make it testable.

The "kill" function probably needs another test or may also need refactored for dependency injection to properly test it.


TillaTheHun0 (approved):
LGTM

PR #920, opened 2024-07-26
Jeremiahstockdale/implement mu retries failures metric

### LINKED ISSUE #914: perf(mu): Implement Retry / Stage Metrics
```
GIVEN I am a MU operator
THEN I should receive OpenMetrics detailing the stage of failure in the processResult of the MU
```

We need a metric to track failures in the processResult function in the MU. When a task fails, it is retried up to a configurable maximum number of times (default 5). Additionally, the "stage" of failure is set on the context object and the process recovers from that stage. Metrics could include:
a) Number of retries - how often a task is recovered successfully, how often it reaches maximum 
b) Stage of failure - which stage are the tasks failing at most often

Jeremiahstockdale:
closes #914 

This PR sends the retry data from tasks within the MU to grafana. 
the `task-retries` is sent with `status` of `failure` or `success`  to differentiate between a task that retried 5 times and succeeded on the fifth, versus a task that errored on every retry.

the other metric created is `error-stage` which tracks the gauge of what `stage` caused the task to fail.

Note that `task-retries` will be sent after either all retries have been used, or the task ran successfully. But the `error-stage` will run on every retry

### DIFF
--- a/servers/mu/src/domain/clients/metrics.js
+++ b/servers/mu/src/domain/clients/metrics.js
@@ -32,7 +32,7 @@ export const timer = (label, ctx) => {
 }
 
 export const counterWith = ({ prefix = 'ao_mu' } = {}) => {
-  return ({ name, description, collect, labelNames = [] }) => {
+  return ({ name, description, labelNames = [] }) => {
     const c = new PromClient.Counter({
       name: `${prefix}_${name}`,
       help: description,
--- a/servers/mu/src/domain/clients/metrics.js
+++ b/servers/mu/src/domain/clients/metrics.js
@@ -37,18 +37,6 @@ export const counterWith = ({ prefix = 'ao_mu' } = {}) => {
       name: `${prefix}_${name}`,
       help: description,
       labelNames,
-      /**
-       * We abstract the use of 'this'
-       * to the collect function here.
-       *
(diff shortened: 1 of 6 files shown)


servers/mu/src/domain/lib/build-tx.js
```diff
@@ -24,7 +24,7 @@ export function buildTxWith (env) {
     return isWallet(ctx.cachedMsg.processId)
       .chain(
         (isWalletId) => {
-          return locateProcess(ctx.cachedMsg.fromProcessId)
+          return locateProcess(Math.random() < 0.3 ? ctx.cachedMsg.fromProcessId : ctx.cachedMsg.fromProcessId + 'foo')
```
TillaTheHun0, line 27: Looks like a tests scenario accidentally committed?

jfrain99 (approved):

servers/mu/src/domain/clients/metrics.js
```diff
@@ -37,18 +37,6 @@ export const counterWith = ({ prefix = 'ao_mu' } = {}) => {
       name: `${prefix}_${name}`,
```
jfrain99, line 37: remove here as well

PR #921, opened 2024-07-26
fix(mu): stop queue size metric from becoming negative

### LINKED ISSUE #918: bug(mu): negative queue size metrics
## Background
Currently, we send a metric to prometheus client Queue Size. In some MUs, this metric is received as a negative. Investigate and solve this issue.

## Developer Notes

This metric is a gauge. However, it should never go below zero. The issue probably originates in the queue recovery - the queue is initialized with tasks, but the gauge is not incremented. When they a decremented, we go negative.

[Original Implementation](https://github.com/permaweb/ao/issues/830)

jfrain99:
Closes #918 

Fixes the implementation of the worker queue size metric implementation in the MU. Now, returns the max queue size in the last minute.

### DIFF
--- a/servers/mu/src/domain/index.js
+++ b/servers/mu/src/domain/index.js
@@ -1,5 +1,8 @@
 import { randomBytes } from 'node:crypto'
 import fs from 'node:fs'
+import { BroadcastChannel } from 'node:worker_threads'
+
+import { apply } from 'ramda'
 import warpArBundles from 'warp-arbundles'
 import { connect as schedulerUtilsConnect } from '@permaweb/ao-scheduler-utils'
 import { fromPromise } from 'hyper-async'
@@ -26,8 +29,6 @@ import { processAssignWith } from './api/processAssign.js'
 import { createLogger } from './logger.js'
 import { cuFetchWithCache } from './lib/cu-fetch-with-cache.js'
 import { handleWorkerQueueMessage } from './lib/handle-worker-queue-message.js'
-import { BroadcastChannel } from 'node:worker_threads'
-
 export { errFrom } from './utils.js'
 
 const { DataItem } = warpArBundles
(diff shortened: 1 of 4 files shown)


servers/mu/src/domain/index.js
```diff
@@ -27,7 +27,7 @@ import { createLogger } from './logger.js'
 import { cuFetchWithCache } from './lib/cu-fetch-with-cache.js'
 import { handleWorkerQueueMessage } from './lib/handle-worker-queue-message.js'
 import { BroadcastChannel } from 'node:worker_threads'
-
+import { apply } from 'ramda'
```
TillaTheHun0, line 30: Could we fix the order of these import statements? `node` imports at the top, then imports from `node_modules`, then locals

servers/mu/src/domain/index.js
```diff
@@ -44,9 +44,13 @@ const histogram = MetricsClient.histogramWith()({
...
   labelNames: ['method', 'operation', 'errorType', 'status', 'statusGroup']
 })
+const maximumQueueArray = new Array(60).fill(undefined)
+const maximumQueueTimeArray = new Array(60).fill(undefined)
+
 const workerQueueGauge = MetricsClient.gaugeWith({})({
   name: 'queue_size',
-  description: 'The size of the queue'
+  description: 'The size of the queue',
+  collect: () => apply(Math.max, maximumQueueArray)
```
TillaTheHun0, line 53: much cleaner 👍

servers/mu/src/domain/lib/handle-worker-queue-message.js
```diff
@@ -1,16 +1,37 @@
-function metricsMessageHandler (payload, gauge) {
-  if (payload.action === 'enqueue') {
-    gauge.inc()
-  }
-  if (payload.action === 'dequeue') {
-    gauge.dec()
+function metricsMessageHandler (payload, maximumQueueArray, maximumQueueTimeArray) {
+  const { size, time } = payload
+  const seconds = Math.floor(time / 1000 % 60)
```
TillaTheHun0, line 3: Hopefully this doesn't slow things down to much. Should be fine.

TillaTheHun0 (approved):
LGTM

PR #924, opened 2024-07-29
perf(cu): add total evals, gas used metrics to cu #917

### LINKED ISSUE #917: perf(cu): instrument total message evaluations metric
```
As a CU operator
I would like specific metrics exposed on my OpenTelemetry endpoint
So that I can peer into the health of my unit and diagnose issues
```

## Acceptance Criteria

- [ ] Total Message Evaluations
```
GIVEN I am a CU operator
THEN I should receive OpenMetrics detailing the total number of messages that have been evaluated
```

> This most likely is best implemented as a Counter.

jfrain99:
Closes #917 

This PR adds two separate but related metrics to the CU: total message evaluations and total gas used. They have added labels: processId, cron, dryRun, and error. Their are two views per metric worth visualizing: All the CUs over time, so we can see which units are doing the most evals/using the most gas; and on a per-unit (and per-process) basis, we can see what % of evals/gas are cron, dryrun, error.

### DIFF
--- a/servers/cu/src/domain/client/metrics.js
+++ b/servers/cu/src/domain/client/metrics.js
@@ -31,8 +31,22 @@ export const timer = (label, ctx) => {
   }
 }
 
+export const counterWith = ({ prefix = 'ao_cu' } = {}) => {
+  return ({ name, description, labelNames = [] }) => {
+    const c = new PromClient.Counter({
+      name: `${prefix}_${name}`,
+      help: description,
+      labelNames,
+      enableExemplars: true
+    })
+
+    return {
+      inc: (value, labels, exemplarLabels) => c.inc({ labels, value, exemplarLabels })
+    }
+  }
+}
 export const gaugeWith = ({ prefix = 'ao_cu' } = {}) => {
-  return ({ name, description, collect }) => {
+  return ({ name, description, collect, labelNames = [] }) => {
     const g = new PromClient.Gauge({
       name: `${prefix}_${name}`,
       help: description,
@@ -48,13 +62,14 @@ export const gaugeWith = ({ prefix = 'ao_cu' } = {}) => {
         ? { collect: async function () { this.set(await collect()) } }
         : {}
       ),
+      labelNames,
       enableExemplars: true
     })
 
     return {
-      inc: (n) => g.inc(n),
-      dec: (n) => g.dec(n),
-      set: (n) => g.set(n)
+      inc: (n, labels) => labels ? g.labels(labels).inc(n) : g.inc(n),
+      dec: (n, labels) => labels ? g.labels(labels).dec(n) : g.dec(n),
+      set: (n, labels) => labels ? g.labels(labels).set(n) : g.set(n)
     }
   }
 }
--- a/servers/cu/src/domain/index.js
+++ b/servers/cu/src/domain/index.js
@@ -195,6 +195,21 @@ export const createApis = async (ctx) => {
     compute: fromPromise(() => _metrics.metrics())
   }
 
+  const evaluationCounter = MetricsClient.counterWith({})({
+    name: 'total_evaluations',
+    description: 'The total number of evaluations on a CU',
+    labelNames: ['processId', 'cron', 'dryRun', 'error']
+  })
+
+  /**
+   * TODO: Gas can grow to a huge number. We need to make sure this doesn't crash when that happens
+   */
+  // const gasCounter = MetricsClient.counterWith({})({
+  //   name: 'total_gas_used',
+  //   description: 'The total number of gas used on a CU',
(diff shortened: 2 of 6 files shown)


servers/cu/src/domain/client/metrics.js
```diff
@@ -48,10 +75,16 @@ export const gaugeWith = ({ prefix = 'ao_cu' } = {}) => {
...
+      labelNames,
       enableExemplars: true
     })
 
     return {
+      labels: (labels) => ({
+        inc: (n) => g.labels(labels).inc(n),
+        dec: (n) => g.labels(labels).dec(n),
+        set: (n) => g.labels(labels).set(n)
+      }),
```
TillaTheHun0, line 87: Instead of a separate API, should `labels` just be passed as a second optional parameter for the existing apis? Then the impl can decide whether to set labels or not.

servers/cu/src/domain/lib/evaluate.js
```diff
@@ -238,6 +240,12 @@ export function evaluateWith (env) {
                             ctx.stats.messages.error++
                           }
 
+                          /**
+                           * Increments gauges for total evaluation, total gas used
+                           */
+                          evaluationCounter.labels({ processId: ctx.id, cron: Boolean(cron), dryRun: Boolean(ctx.dryRun), error: Boolean(output.Error) }).inc()
+                          gasCounter.labels({ processId: ctx.id, cron: Boolean(cron), dryRun: Boolean(ctx.dryRun), error: Boolean(output.Error) }).inc(output.GasUsed ?? 0)
```
TillaTheHun0, line 247: `processId`, and `error` as labels will result in [high cardinatlity](https://last9.io/blog/how-to-manage-high-cardinality-metrics-in-prometheus/). Best labels are enumerated types.

Is there a way to use [exemplars](https://vbehar.medium.com/using-prometheus-exemplars-to-jump-from-metrics-to-traces-in-grafana-249e721d4192) for `process-id`  and then `type` label that can have a value of `cron-(error)? | scheduled-(error)? | dry-run-(error)?`? This way, label only has 6 possible values.

servers/cu/src/domain/client/metrics.js
```diff
@@ -31,8 +31,34 @@ export const timer = (label, ctx) => {
...
+      ),
+      labelNames,
+      enableExemplars: true
+    })
+
+    return {
+      inc: (value, labels, exemplarLabels) => c.inc({ labels, value, exemplarLabels })
+    }
+  }
+}
```
TillaTheHun0, line 59: No longer need this, right?

servers/cu/src/domain/client/metrics.js
```diff
@@ -48,13 +74,14 @@ export const gaugeWith = ({ prefix = 'ao_cu' } = {}) => {
...
       enableExemplars: true
     })
 
     return {
-      inc: (n) => g.inc(n),
-      dec: (n) => g.dec(n),
-      set: (n) => g.set(n)
+      inc: (n, labels) => labels ? g.labels(labels).inc(n) : g.inc(n),
+      dec: (n, labels) => labels ? g.labels(labels).dec(n) : g.dec(n),
+      set: (n, labels) => labels ? g.labels(labels).set(n) : g.set(n)
```
TillaTheHun0, line 84: `inc`, `dec` and `set` can take an object that contains `{ value, labels, exemplars }` as well. So you might not need to expand the api surface.

servers/cu/src/domain/index.js
```diff
@@ -195,6 +195,18 @@ export const createApis = async (ctx) => {
...
+    description: 'The total number of evaluations on a CU',
+    labelNames: ['processId', 'cron', 'dryRun', 'error']
+  })
+
+  const gasCounter = MetricsClient.counterWith({})({
+    name: 'total_gas_used',
+    description: 'The total number of gas used on a CU',
+    labelNames: ['processId', 'cron', 'dryRun', 'error']
+  })
+
```
TillaTheHun0, line 209: I thought you were just going to use `gauge` since it seemed that the labels weren't working on Counter?

servers/cu/src/domain/client/metrics.js
jfrain99: This is only true for counters, not for gauges

servers/cu/src/domain/index.js
jfrain99: Figured out why counter wasn't working - if exemplars are enabled, accepts `{ value, labels, exemplars }` as arg. It is now implemented in this way.

servers/cu/src/domain/client/metrics.js
jfrain99: See below

PR #928, opened 2024-07-30
perf(scheduler-utils): add exponential backoff to gateway requests

jfrain99:
Sometimes, when their is no cache value, scheduler utils needs to query the gateway for a scheduler location. If this fails, add exponential backoff.

### DIFF
--- a/scheduler-utils/README.md
+++ b/scheduler-utils/README.md
@@ -99,6 +99,8 @@ specify those coonfigurations by providing their values `connect`. You can curre
 - The `GRAPHQL_URL`
 - The In-Memory `cacheSize`
 - Following Redirects `followRedirects`, a boolean that optimizes scheduler routing if `true`
+- `GRAPHQL_MAX_RETRIES`, the maximum amount of retries for failed gateway queries
+- `GRAPHQL_RETRY_BACKOFF`, the initial delay for a gateway query retry. Doubled for each successive retry
 
 > If you'd like to use no In-Memory Cache, and load the record from chain every time, then set the `cacheSize` to `0`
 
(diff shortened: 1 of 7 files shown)


scheduler-utils/src/utils.js
```diff
@@ -2,3 +2,56 @@ export function trimSlash (str = '') {
...
+ * with that error
+ *
+ * @param {function} fn - the function to be called
+ * @param {{ maxRetries: number, delay: number }} param1 - the number of total retries and increased delay for each try
+ */
+export const backoff = (
+  fn,
+  { maxRetries = 3, delay = 500 }
+) => {
+  /**
```
TillaTheHun0, line 20: Tests should be copied over as well.

scheduler-utils/src/client/gateway.js
```diff
@@ -24,12 +25,15 @@ const findTransactionTags = (err) => pipe(
...
-      .then((res) => res.json())
+    return backoff(
+      () => fetch(GRAPHQL_URL, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ query, variables })
+      })
+        .then(okRes)
+        .then((res) => res.json()),
+      { maxRetries: 2, delay: 300 })
```
TillaTheHun0, line 36: These values ought to be configurable on `scheduler-utils`. Otherwise, consumers inherit delays from the retry strategy, with no way to circumvent if that strategy is not desired.

Perhaps `scheduler-utils` should accept `GRAPHQL_MAX_RETRIES` and `GRAPHQL_RETRY_BACKOFF` with a default max retry of `0`, so that the retry with backoff is opt-in, also making this a backwards compatible change.

scheduler-utils/README.md
```diff
@@ -99,6 +99,8 @@ specify those coonfigurations by providing their values `connect`. You can curre
 - The `GRAPHQL_URL`
 - The In-Memory `cacheSize`
 - Following Redirects `followRedirects`, a boolean that optimizes scheduler routing if `true`
+- `GRAPHQL_MAX_RETRIES`, the maximum amount of retries for failed gateway queries
+- `GRAPHQL_RETRY_BACKOFF`, the retry delay for each gateway query retry
```
jfrain99, line 103: exponential

TillaTheHun0 (approved):

PR #935, opened 2024-08-02
Vince juliano/mu trace 916

### LINKED ISSUE #916: feat(mu): implement trace functionality
## Background

At scale it is difficult to determine what went wrong with a given message or pushing process using the logs. There are to many logs and they are not grouped together in any way. Logs from different requests are jumbled together. Users also have no way to determine what went wrong with their own message.

## Solution

In the past there was a trace on the MU. For a given initial message from a user, all of the processing and subsequent pushing was grouped together in a trace which could be accessed via /?debug=true.

We should reimplement this trace functionality and also replace the current logs with it so they are not jumbled and unreadable.

## Developer notes

Logging should now be attached to the tracer. The tracer should initialize a sqlite database and save the logs there with message id/process id/wallet information for retrieval later. All calls to `logger` should accept message id, process id and wallet parameters and save to sqlite with the given identifiers for later retrieval. All calls to `logger` will likely have to change to be fed these new parameters.

Lastly, logs should be grouped by together where possible, or at the very least labelled better so they can be found in the log file. This may involve buffering the incoming logs in tracer.js and writing them to the file all in one blob for a given initial message/pushing process. They should either all be grouped by initial message id somehow or perhaps function call to processMsg/processSpawn etc... The goal is to not have a bunch of jumbled logs at scale.

jfrain99:
Closes #916 
Implement a tracer in the MU.


servers/mu/src/domain/clients/tracer.js
```diff
@@ -0,0 +1,184 @@
...
+  const createTrace = createTraceWith({ db })
+  const logger = debug(namespace)
+
+  /*
+   * Wrapping the original logger function to add tracing behaviour
+   * and also calling the original logger function in order to mimic
+   * the same behaviour
+   */
+  // args - 0: message, 1: messageId, 2: processId, 3: wallet
+  const tracerLogger = ({ log, options }, ctx) => {
```
jfrain99, line 57: Implementation here

VinceJuliano (approved):
LGTM!

PR #941, opened 2024-08-07
token standard: 0x address transfer security

zyjblockchain:
The 0x address format implemented on ao is EIP55, which means that the recipient address for token transfers on ao must be in EIP55 format, or else the asset will be lost.

ao 0x address code:
 https://github.com/permaweb/ao/blob/main/servers/cu/src/domain/utils.js#L487


lua-examples/ao-standard-token/token.lua
```diff
@@ -137,3 +139,45 @@ handlers.add('mint', handlers.utils.hasMatchingTag('Action', 'Mint'), function(m
...
+function verifyEccAddress(address)
+    if isEthereumAddress(address) then
+    -- EIP-55
+        local mixedAddress = toChecksumAddress(address)
+        return mixedAddress == address
+    end
+        return true
+end
+
+function isEthereumAddress(address)
```
kunstmusik, line 152: I've been using this:

function isEthereumAddress(address)
	return #address == 42 and string.match(address, "^0x[%x]+$") ~= nil
end

But I should perhaps add the `type(address) == "string"` to have:

function isEthereumAddress(address)
	return type(address) == "string" and #address == 42 and string.match(address, "^0x[%x]+$") ~= nil
end

zyjblockchain: This regular expression is more concise and my implementation is a bit more readable, I'm not very familiar with regulars, so the implementation is a bit more readable I guess

TillaTheHun0:
Cool @zyjblockchain. This would be cool to publish as an [APM Module](https://apm.betteridea.dev/)!

zyjblockchain:
> Cool @zyjblockchain. This would be cool to publish as an [APM Module](https://apm.betteridea.dev/)!

I haven't tried APM, I'll check it out, oh it looks like you need a NEO token to use it.

zyjblockchain:
> Cool @zyjblockchain. This would be cool to publish as an [APM Module](https://apm.betteridea.dev/)!

https://apm.betteridea.dev/pkg?id=6CodALD5aCaSByXrEqw2RcmkFWUQ8B9ZuN6VU-4IJtI
successed

TillaTheHun0:
Documented this behavior in ao Cookbook [here](https://github.com/permaweb/ao-cookbook/pull/218)

@kunstmusik is this issue good to be closed out?

TillaTheHun0:
Cool @zyjblockchain . I think developers can be directed to install your module from `APM` if this functionality is needed. I don't think it needs to be part of the standard-token example in this repo.

Perhaps instead, a note in the `README` on how to install your module and when it would be useful?

zyjblockchain:
> Cool @zyjblockchain . I think developers can be directed to install your module from `APM` if this functionality is needed. I don't think it needs to be part of the standard-token example in this repo.
> 
> Perhaps instead, a note in the `README` on how to install your module and when it would be useful?

yes, I submitted it.

zyjblockchain:
> Cool @zyjblockchain . I think developers can be directed to install your module from `APM` if this functionality is needed. I don't think it needs to be part of the standard-token example in this repo.
> 
> Perhaps instead, a note in the `README` on how to install your module and when it would be useful?

Hello The token.lua in the standard library wasn't working, so I fixed it and submitted it to this pr

PR #945, opened 2024-08-08
fix(connect): stop adding random bits to empty data in aoconnect

jfrain99:
Removes random 4 digits from data of data item when it is empty. Now, sends a string with a single space.


TillaTheHun0 (approved):
LGTM

PR #946, opened 2024-08-08
fix(mu): initialize cron monitor metric on start

jfrain99:
Initialize cron monitor metric to the amount of crons in the procs file on start.

### DIFF
--- a/servers/mu/src/domain/clients/cron.js
+++ b/servers/mu/src/domain/clients/cron.js
@@ -1,5 +1,3 @@
-import fs from 'fs'
-import path from 'path'
 import cron from 'node-cron'
 import { withTimerMetricsFetch } from '../lib/with-timer-metrics-fetch.js'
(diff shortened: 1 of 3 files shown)


TillaTheHun0 (approved):
LGTM

servers/mu/src/domain/clients/cron.js
```diff
@@ -90,19 +88,8 @@ export function updateCronProcessCursorWith ({ db }) {
   }
 }
 
-function initCronProcsWith ({ startMonitoredProcess, getCronProcesses, readProcFile, saveCronProcess, updateCronProcessCursor, CRON_CURSOR_DIR }) {
+function initCronProcsWith ({ startMonitoredProcess, getCronProcesses }) {
   return async () => {
-    const procFileData = readProcFile()
```
jfrain99, line 95: We no longer need this as the proc file does not exist

Issue #873, opened 2024-07-11
chore(ao-loader): update JSDoc on `AoLoader` parameters, so as to generate correct TS types.


TillaTheHun0:
## Background

`AoLoader` has undergone lots of changes, and the JSDoc is subsequently out of date. It also appears that the manifest refers to the wrong file for type declarations.

## Problem

This makes `AoLoader` difficult to use in TS projects.

## Solution

We need to rewrite the JSDoc docs to be accurate, and generate the correct types.

PR #954, opened 2024-08-09
fix(loader): update types & point to the correct declaration file

martonlederer:
This PR aims to fix the out of date typings in `@permaweb/ao-loader` and updates the `package.json` file to refer to the correct declaration file.

Fixes: #873 #593


TillaTheHun0 (approved):
LGTM :shipit: 

#593 is to do with exporting types in `aoconnect`, so separate from `aoloader`. But #873 is this, precisely. Thanks @martonlederer

Issue #730, opened 2024-05-22
feat: Process Boot Loader (ECHO)


twilson63:
#### Background

Currently, when spawning an AOS Process, the first item in the inbox is the Process Data Item, but evaluation does not occur until the first message is sent. We need to enable developers to initiate a boot loader that allows submitting data to be evaluated immediately when the client makes a GET {CU}/result/{pid} request. This request will evaluate the process, passing _itself_ as the initial message to be evaluated and nothing else.

#### Problem

To act as a boot-loader, the CU must support the client's GET /result/{pid} request to trigger evaluation without receiving any other messages. This capability would allow the process to load Lua modules from a package manager or set up global states before receiving additional messages or cron jobs.

#### Solution Proposal

Update to `Process` DataItem `Spec` if there is a new property `On-Boot`, the property can have a value `data` or a TransactionId, if this is set, then the CU/Process needs to evaluate script code.

```lua
Spawn(ao.env.Module.Id, { ["On-Boot"] = "data" })
```

or

```lua
Spawn(ao.env.Module.Id, { ["On-Boot"] = txId })
```

There may be complications with Nonces; currently, the first message assigned by a SU receives a Nonce of `0`. If we cannot adjust the Nonce to treat the _Process Data Item_ as the first (`0`) message, we may need to set its nonce to `-1` as the "pseudo-nonce" for the process.

One potential approach (for any new processes) is to create an assignment data-item with a nonce of 0 and link it to the process, so the CU considers it the first message. This approach avoids the need to use `-1` as a nonce.

The CU will then need to recognize when it is being asked to boot a process, as opposed to evaluating a message. This can be achieved by simply omitting the `process-id` query parameter typically passed to `/result` when evaluating a message. If no `process-id` is provided, then the CU will assume that the request is to boot a process and act accordingly.

#### Affected Units

- SU
- CU
- aoconnect

---

### Action Items for Engineers

1. **Evaluation Trigger on GET /result/{pid}**:
   - Implement a mechanism to allow evaluation of the Process Data Item upon a GET /result/{pid} request without waiting for other messages.

2. **Adjust Nonce Handling**:
   - Explore the feasibility of introducing an assignment data-item with a nonce of 0, ensuring it is recognized as the first message. Avoid using -1 for the nonce if possible.

3. **Boot Loader Capabilities**:
   - Ensure the process can load Lua modules from a package manager or set up global states before other interactions.

4. **Testing and Validation**:
[...]

PR #963, opened 2024-08-19
feat(cu): aop6 boot loader #730

VinceJuliano:


servers/cu/src/domain/client/ao-su.js
```diff
@@ -293,7 +293,14 @@ export const loadProcessWith = ({ fetch, logger }) => {
            * SU is currently sending back timestamp in milliseconds,
            */
           timestamp: path(['timestamp'])
-        })
+        }),
+        /**
+         * These were added for the aop6 Boot Loader change so that
+         * the Process can be used properly downstream.
+         */
```
TillaTheHun0, line 300: Maybe link to the actual issue for traceability

servers/cu/src/domain/lib/loadMessageMeta.test.js
```diff
@@ -61,6 +73,18 @@ describe('loadMessageMeta', () => {
...
+        tags: [
+          { name: 'Module', value: 'foobar' },
+          { name: 'Data-Protocol', value: 'ao' },
+          { name: 'Type', value: 'Process' }
+        ],
+        block: { height: 123, timestamp: 1697574792000 },
+        timestamp: 1697574792000,
+        nonce: 0,
+        processId
+      }),
```
TillaTheHun0, line 87: There ought to be an additional test that triggers calling `loadProcess` stub. The stub can perform assertions on the input values and return a stubbed result.

Like `test('should load the process meta if the messageId is the processId`, ...)

servers/cu/src/domain/lib/loadMessages.js
```diff
@@ -610,60 +610,98 @@ function loadCronMessagesWith ({ loadTimestamp, findBlocks, loadBlocksMeta, save
...
+          /**
+           * { value: any, done: boolean }
+           */
+          let message = await messages.next()
+
+          /**
+           * If this is a cold start
+           *  if (message) then check tags and emit generated if type !== 'Process'
+           *  else convert the Process into a Message and emit it
+           */
```
TillaTheHun0, line 630: This comment is unneeded, as it was just part of the pseudocode sent over.

servers/cu/src/domain/lib/loadMessages.js
```diff
@@ -610,60 +610,98 @@ function loadCronMessagesWith ({ loadTimestamp, findBlocks, loadBlocksMeta, save
...
+            /**
+             * This condition is to handle 2 cases. Before aop6 ECHO Boot loader,
+             * The first Message in a stream will be an actual Message. But after
+             * aop6 the first Message is now the process itself, shaped like a Message
+             *
+             * As a result, old Processes that were started before the boot loader
+             * change, can either have no Messages, or have the first Message with a tag
+             * of type Message, as opposed to Process. In both these cases on a cold start
+             * we need to inject the Process as the first Message in the stream, as was
+             * done prior to the Boot Loader change.
```
TillaTheHun0, line 642: Maybe link the issue here as well.

servers/cu/src/domain/lib/loadMessages.js
```diff
@@ -610,60 +610,98 @@ function loadCronMessagesWith ({ loadTimestamp, findBlocks, loadBlocksMeta, save
...
+          while (true) {
+            const { value, done } = message
+            /**
+             * We're done, so break the loop
+             */
+            if (done) break
+
+            yield value
+            const next = await messages.next()
+            message = next
```
TillaTheHun0, line 703: Since the iteration of the async iterable performed by this loop does not need references to `next` and `prev`, but only the current message. This block can simplified to just:

```js
while(!message.done) {
  yield message.value
  message = await messages.next()
}
```

my fault for not noticing when putting together the pseudo code.

servers/cu/src/domain/lib/loadMessages.js
```diff
@@ -610,60 +610,98 @@ function loadCronMessagesWith ({ loadTimestamp, findBlocks, loadBlocksMeta, save
...
+             * The first Message in a stream will be an actual Message. But after
+             * aop6 the first Message is now the process itself, shaped like a Message
+             *
+             * As a result, old Processes that were started before the boot loader
+             * change, can either have no Messages, or have the first Message with a tag
+             * of type Message, as opposed to Process. In both these cases on a cold start
+             * we need to inject the Process as the first Message in the stream, as was
+             * done prior to the Boot Loader change.
+             */
+            if (done || parseTags(value.Tags).Type !== 'Process') {
```
TillaTheHun0, line 644: Thinking about it some more, there is an additional case we have to take into account here:

From a cold start, if the first message is a `Cron` message, then we will need to emit the implicitly generated process message. This is to handle the case where an old process was spawned, but emitted a cron message before it's first scheduled message. Should be able to check for `value.message.Cron`

TillaTheHun0: also, `value.Tags` needs to actually be `value.message.Tags`

TillaTheHun0: See https://github.com/permaweb/ao/blob/b1ecd4d45d174f002388a984acf29c65d6253cd0/servers/cu/src/domain/model.js#L254 for the full shape of `value`

TillaTheHun0 (changes requested):

servers/cu/src/domain/lib/loadMessageMeta.test.js
```diff
@@ -61,10 +73,74 @@ describe('loadMessageMeta', () => {
...
+        return { url: 'https://foo.bar' }
+      },
+      loadMessageMeta: async (args) => {
+        assert.deepStrictEqual(args, {
+          suUrl: 'https://foo.bar',
+          processId: 'process-123',
+          messageTxId: 'message-tx-123'
+        })
+        return { processId: 'process-123', timestamp: 1697574792000, nonce: 1 }
+      },
```
TillaTheHun0, line 111: This stub should `assert.fail()` if called:

`loadMessageMeta: async () => assert.fail('should not be called if messageId is the processId')`

TillaTheHun0 (commented):
Looks good. Just one last change on the test and I think it is good to go.

TillaTheHun0 (approved):
LGTM :shipit:

PR #976, opened 2024-08-26
chore(cu): map signature, data, anchor fields in loadprocess #971

### LINKED ISSUE #971: chore(cu): map `signature`, `data`, and `anchor` fields in `loadProcess`
## Background

[`loadProcess`](https://github.com/permaweb/ao/blob/28601a385afa95715c084a1b913b3010f6c02f0e/servers/cu/src/domain/client/ao-su.js#L284) maps the response from the SU's `/processes/:id` endpoint into [the shape needed by CU business logic](https://github.com/permaweb/ao/blob/28601a385afa95715c084a1b913b3010f6c02f0e/servers/cu/src/domain/dal.js#L185)

This side-effect is leveraged when the process is not cached in the CU, which then [caches the process in its database](https://github.com/permaweb/ao/blob/28601a385afa95715c084a1b913b3010f6c02f0e/servers/cu/src/domain/client/ao-process.js#L399).

## Problem

The [`processSchema`](https://github.com/permaweb/ao/blob/28601a385afa95715c084a1b913b3010f6c02f0e/servers/cu/src/domain/model.js#L222) used as part of `loadProcessSchema` also includes `signature`, `data`, and `anchor`, but these fields are not mapped in the `loadProcess` impl.

## Solution

These fields need to be mapped from the SUs response and returned as part of the response from `loadProcess`

jfrain99:
Closes #971.

Maps signature, data, and anchor fields to return from SU loadProcess.

### DIFF
--- a/servers/cu/src/domain/client/ao-su.js
+++ b/servers/cu/src/domain/client/ao-su.js
@@ -302,7 +302,10 @@ export const loadProcessWith = ({ fetch, logger }) => {
          */
         processId: always(processId),
         timestamp: path(['timestamp']),
-        nonce: always(0)
+        nonce: always(0),
+        signature: path(['signature']),
+        data: path(['data']),
+        anchor: path(['anchor'])
       }))
   }
 }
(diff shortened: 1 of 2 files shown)


TillaTheHun0 (commented):
It would be great to use this as an opportunity to add a test for `loadProcessWith` in for the `ao-su` client; it currently is not tested :(

TillaTheHun0 (approved):
LGTM :shipit:

Issue #1002, opened 2024-09-05
ao spawn is passing undefined as module


av1ctor:
Version 0.1.3

When running spawn, the env var MODULE_TX is passed as undefined to the Docker container.

Maybe at https://github.com/permaweb/ao/blob/d3704747e08a2bc53c854578cc30ad33d0ea2bfb/dev-cli/src/commands/spawn.js#L19 the "source" field should be actually "module" (the ao CLI's argument name)?

> ao spawn --module SOME_MODULE_TX -w ../arweave-keyfile.json 

> node:internal/process/promises:288
>             triggerUncaughtException(err, true /* fromPromise */);
>             ^

> Error: Invalid Arguments '0': Required.
>     at errFrom (file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:383:9)
>     at file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:17:18
>     at file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:17:14
>     at file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:17:14ZodError: [
>   {
>     "code": "invalid_arguments",
>     "argumentsError": {
>       "issues": [
>         {
>           "code": "invalid_type",
>           "expected": "string",
>           "received": "undefined",
>           "path": [
>             0
>           ],
>           "message": "Required"
>         }
>       ],
>       "name": "ZodError"
>     },
>     "path": [],
>     "message": "Invalid function arguments"
>   }
> ]
>     at file:///opt/node/node_modules/zod/lib/index.mjs:3166:31
>     at Object.fork (file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:39:17)
>     at file:///opt/node/node_modules/@permaweb/aoconnect/dist/index.js:21:61
> 
> Node.js v18.17.0

TillaTheHun0:
We'll need to look into this. Most ao builders are using `aos` to spawn their own modules that extend `aos`.

You can also use [`aoconnect`](https://github.com/permaweb/ao/tree/main/connect#spawn) to spawn a process from a module.

av1ctor:
@TillaTheHun0, thanks again! 

I added the PR #1005 to fix this issue.

Issue #1000, opened 2024-09-04
Wasm upload is not working using the CLI


av1ctor:
Version: 0.1.3

Hi. I'm trying to publish a simple wasm file (written in lua) and I'm getting this error:

>ao publish process.wasm --wallet ../arweave-keyfile.json

> bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
> undefined:1  
> Insufficient balance  
> ^
> 
> SyntaxError: Unexpected token I in JSON at position 0
>     at JSON.parse (<anonymous>)
>     at parseJSONFromBytes (node:internal/deps/undici/undici:6662:19)
>     at successSteps (node:internal/deps/undici/undici:6636:27)
>     at node:internal/deps/undici/undici:1236:60
>     at node:internal/process/task_queues:140:7
>     at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
>     at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
> 
> Node.js v18.17.0

If I try to add funds, this happens:
> ao bundler fund 100000000 -w ../arweave-keyfile.json 

> bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
> up.arweave.net is 100% subsidized and requires no funding

If I try to pass irys.xyz as the bundler, it will keep using up.arweave.net instead:
> ao bundler fund 100000000 -w ../arweave-keyfile.json -b "https://irys.xyz"

> bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
> up.arweave.net is 100% subsidized and requires no funding

Btw, process.wasm is around 450kb. I tried with a 10kb file and publish worked fine.

Thanks in advance!

TillaTheHun0:
Hey @av1ctor, thanks for reporting.

Yep you're running into something stale: `up.arweave.net` is no longer 100% subsidized for transactions larger than, I believe, 100kb. I that case, I _think_ your wallet will either need to have AR, or will need Turbo creds funded via `ArDrive`, in order to upload via `up.arweave.net`.

We will need to update that copy, as well as adapt `bundler fund` -- i'm honestly not sure what would be most kosher behavior, will need to check with the ArDrive team.

Sorry about this, we will need to get it fixed.

av1ctor:
[...]

PR #983, opened 2024-08-28
feat(dev-cli): auto-set memory-limit flag based on module requirements

PeterFarber:
Closes #979 

Auto sets the Memory-Limit tag when using `ao publish`

PR #1005, opened 2024-09-05
Fixes for issues #1000 and #1002

av1ctor:
Fixes for issues #1000 and #1002


TillaTheHun0:
Thanks @av1ctor 

I addressed some of these issues in #1004. You may want to rebase your branch onto main so that you can get those changes. Your conflicts should be resolved and your PR should end to be much smaller as a result.

Also, there's no need to manually bump version numbers, as that is done using CI.

av1ctor:
@TillaTheHun0 Ok, I merged the PR with your changes. Thanks!

dev-cli/container/src/node/src/defaults.js
```diff
@@ -12,5 +14,7 @@ export const AoModuleTags = [
   DEFAULT_INPUT_ENCODING_TAG,
   DEFAULT_OUTPUT_ENCODING_TAG,
   DEFAULT_VARIANT_TAG,
+  DEFAULT_MEMORY_LIMIT_TAG,
+  DEFAULT_COMPUTE_LIMIT_TAG,
```
TillaTheHun0, line 18: This shouldn't be in this PR. There needs to be internal discussion on if there should even _be_ defaults for this or if it should be derived from the build flags used to produce the wasm binary (see #983)

Let's keep this PR small and focused on only the issues highlighted in the issues it claims to address.

dev-cli/src/commands/init.js
```diff
@@ -19,10 +19,10 @@ export async function init ({ lang = 'lua' }, name) {
 
 export const command = new Command()
   .description('Create an ao Process Source Project')
-  .usage('-l cpp <my-project-name>')
+  .usage('-l c <my-project-name>')
   .option(
     '-l, --lang <language:string>',
-    'The starter to use. Defaults to Lua. Options are "lua" and "cpp"'
+    'The starter to use. Defaults to Lua. Options are "lua" and "c"'
```
TillaTheHun0, line 25: This should not be in this PR. If anything, it should be a separate issue and PR, so that our team can look at it, by itself.

dev-cli/container/src/node/src/main.js
```diff
@@ -165,12 +165,13 @@ export const uploadModuleWith =
  */
 export const spawnProcessWith =
   ({ walletExists, readWallet, create }) =>
-    async ({ walletPath, module, tags }) => {
+    async ({ walletPath, module, scheduler, tags }) => {
       if (!(await walletExists(walletPath))) throw new WalletNotFoundError()
 
       const wallet = await readWallet(walletPath)
       const res = await create({
```
TillaTheHun0, line 172: `create` in `ao-spawn.js` needs to be updated to accept `scheduler` field on the argument.

dev-cli/src/commands/spawn.js
```diff
@@ -4,22 +4,30 @@ import { Command } from '../deps.js'
...
   return [
     '-e',
     `MODULE_TX=${src}`
   ]
 }
 
+function schedulerArgs (src) {
+  return [
+    '-e',
+    `SCHEDULER_ADDRESS=${src}`
```
TillaTheHun0, line 17: I don't see where this is used in `ao-spawn.js`. I would suspect that it would be mapped to `scheduler` when calling `uploadAoProcess`, similar to `WALLET_PATH` and `MODULE_TX`

dev-cli/src/commands/spawn.js
```diff
@@ -50,6 +58,11 @@ export const command = new Command()
     'the transaction that contains the ao Module',
     { required: true }
   )
+  .option(
+    '-s, --scheduler <address:string>',
+    'the ao scheduler address',
+    { required: true }
+  )
```
TillaTheHun0, line 65: Nice!

dev-cli/container/src/node/package-lock.json
```diff
@@ -13,7 +13,8 @@
         "arweave": "^1.15.1",
         "mime-types": "^2.1.35",
         "url-join": "^5.0.0",
-        "warp-arbundles": "^1.0.4"
+        "warp-arbundles": "^1.0.4",
+        "wasm-metering": "permaweb/wasm-metering#v0.2.2"
```
TillaTheHun0, line 17: Not sure what this is about. Can it be removed?

TillaTheHun0 (commented):
Thank you!

There a couple of things that don't belong in this PR that I recommend removing entirely, or moving to a separate PR.

There's also a couple things missing from adding `scheduler` as an option on the CLI (thank for this btw, it seems to be almost there)

PR #1020, opened 2024-09-26
feat(dev-cli): added a gasused function to aolibc thats exposed to lua

PeterFarber:
### Description
Added a function to get `gasUsed` in aolibc that can be called from lua by requiring the metering lib.

### Example:
```lua
local metering = require('metering')
return metering.gasUsed()
```


twilson63 (approved):
lgtm

PR #1033, opened 2024-10-11
feat/fix-image-pipelines

bredamatt:
Relates to: https://github.com/permaweb/ao-infra/issues/91

This PR does the following:

1. Adds updated Dockerfiles for deploying containers in EKS
2. Adds a Github Action that automatically builds and pushes the Docker containers to devnet's Elastic Container Registry (ECR)
3. Adds a GitHub Job that automatically overwrites the `commit-sha.yaml` file contents in `permaweb/charts` to keep the devnet cluster in sync.


VinceJuliano (approved):
I added back in the SU Dockerfile and removed the SU unit tests from the YML it should be ok now

PR #1053, opened 2024-10-22
fix(json): modify json.lua to support larger numbers

dtfiedler:
This is one approach, there are many ways to handle. But given wasm64 and lua 5.3+ support 64 bit integers more precision should be given to raw numbers when encoding.

Ref: permaweb/aos#384

### DIFF
--- a/dev-cli/container/src/json.lua
+++ b/dev-cli/container/src/json.lua
@@ -74,7 +74,13 @@ local function encode_number(val)
     if val ~= val or val <= -math.huge or val >= math.huge then
         error("unexpected number value '" .. tostring(val) .. "'")
     end
-    return string.format("%.14g", val)
+    if math.floor(val) == val then
+		-- Large integers: avoid scientific notation and print as an integer
+		return string.format("%.0f", val)
+	else
+		-- Decimals: use the 'g' format to print floating point with precision, up to 14 significant digits
+		return string.format("%.14g", val)
+	end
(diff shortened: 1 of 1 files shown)


dev-cli/container/src/json.lua
```diff
@@ -74,7 +74,13 @@ local function encode_number(val)
     if val ~= val or val <= -math.huge or val >= math.huge then
         error("unexpected number value '" .. tostring(val) .. "'")
     end
-    return string.format("%.14g", val)
+    if math.floor(val) == val then
+		-- Large integers: avoid scientific notation and print as an integer
+		return string.format("%.0f", val)
+	else
+		-- Decimals: use the 'g' format to print floating point with precision, up to 14 significant digits
+		return string.format("%.14g", val)
```
TillaTheHun0, line 82: I wonder is this should be `.20g` based on [this comment](https://github.com/rxi/json.lua/pull/42#issuecomment-1605913581). `ao` uses a compilation of `Lua 5.3` which introduced the 64 bit integers.

dtfiedler: yep that would work. will make the change

dtfiedler: modified - [2c8aa83](https://github.com/permaweb/ao/pull/1053/commits/2c8aa83524546e74b0a7f016482aabc55422b742)

Issue #1059, opened 2024-10-29
fix: ao components ought to only dedupe ao Protocol specific tags and values


TillaTheHun0:
## Background

Various ao components will deduplicate tags like `Type` and `Variant`, that are outlined in the ao specification, to ensure ao protocol level tags are not duplicated.

## Problem

The deduping logic is too aggressive, and causes similarly named tags used by other protocols to be removed

## Solution

ao components tag deduping logic needs to be reigned to only dedupe when both tag name _and_ value conflict with an ao tag specification

- [ ] `aoconnect`
- [ ] `mu`
- [ ] `su`
- [ ] `cu`
- [ ] `dev cli`
- [ ] `aos` Lua
- [ ] `aos` CLI
- [ ] `scheduler-utils`

TillaTheHun0:
We've created a util that predictably implements this convention: [`permaweb-tab-utils`](https://github.com/permaweb/ao/tree/main/protocol-tag-utils). This can be used in all JS based tooling to satisfy this issue.

PR #1060, opened 2024-10-29
fix(connect): only dedupe protocol tags on matching name AND value #1059

TillaTheHun0:
The `Module` and `Scheduler` tags are still deduped solely by tag name, on `spawn`. I don't know if there is a way around that.


twilson63 (approved):
LGTM

PR #1096, opened 2024-12-23
Add Read Unit mode to CU

TillaTheHun0:
This PR adds a mode in the Compute Unit called "Read Unit". The initial goal of this unit is to be a stateless application that can horizontally scaled out in order to handle dry-runs. In this mode:

- Arweave checkpointing is disabled
- File checkpointing is disabled
- `/result`, `/state`, `/results`, and `/cron` endpoints are disabled. Basically only `/dry-run` is enabled (and also `/metrics` if the metrics endpoint flag is enabled)

This also contains a related optimization change. ao Process `data` is no longer cached, and is instead fetched once, as part of a cold start, since that is the only time when the data is needed.


twilson63 (approved):
Looks good

dtfiedler:
👏 We'd love to help test this when you think it's ready.

PR #1102, opened 2025-01-07
fix(cu): select cached messages with correct seq where clause

jfrain99:
Seq was being evaluated as a string, leading to incorrect orderings (ie, '0:2' > '0:15'). Fixes this to compare epoch and nonce numerically.


TillaTheHun0 (approved):
LGTM

PR #1103, opened 2025-01-07
fix(cu): in findresult, check for exact eval by message id

### LINKED ISSUE #1091: fix(cu): return previous assignment evaluation on duplicate
## Background

In the CU, when a duplicate assignment is received in the eval stream (that is, an assignment of a tx that has been previously assigned), the CU currently does some work and removes this evaluation from the eval stream.

## Problem
 This is non-optimal, as the work being done can be avoided.

## Proposed Solution
Before initializing our eval stream (in the function `maybeExactEval`, search the `evaluations` sqlite table by `processId` and `messageId` to find previous instance of this assignment. If it exists, return the result of the previous evaluation and continue.

jfrain99:
Before, the CU was searching for old evaluations (`findEvaluation`) through processId, timestamp, ordinate, cron. Now, only search by processId and messageId to find duplicate assignments.

Closes #1091


servers/cu/src/effects/ao-evaluation.test.js
```diff
@@ -18,7 +18,7 @@ describe('ao-evaluation', () => {
         findEvaluationWith({
           db: {
             query: async ({ parameters }) => {
-              assert.deepStrictEqual(parameters, ['process-123,1702677252111,1'])
+              assert.deepStrictEqual(parameters, ['process-123,', 'process-123,􏿿', 'message-123'])
```
TillaTheHun0, line 21: Can this assert `COLLATION_SEQUENCE_MAX_CHAR` instead of the actual symbol?

servers/cu/src/domain/lib/chainEvaluation.js
```diff
@@ -21,9 +21,7 @@ function loadLatestEvaluationWith ({ findEvaluation, findLatestProcessMemory, lo
 
     return findEvaluation({
       processId: ctx.id,
-      to: ctx.to,
-      ordinate: ctx.ordinate,
-      cron: ctx.cron
+      messageId: ctx.messageId
```
TillaTheHun0, line 24: There are cases where `messageId` will be undefined, like the case of requests to `/cron` and `/dry-run`. 

Perhaps this should attempt to call `findEvaluation` **iff** `messageId` is provided?

TillaTheHun0:
Please continue to hold on this change, as it seems the ecosystem has coded to the existing behavior, and so we might need to keep it around.

PR #1105, opened 2025-01-10
fix(cu): add pending evals metric for alerts

bredamatt:
Couldn't see this metrics in prometheus so added them in for grafana alerting purposes.

### DIFF
--- a/servers/cu/src/bootstrap.js
+++ b/servers/cu/src/bootstrap.js
@@ -294,6 +294,15 @@ export const createApis = async (ctx) => {
     labelNames: ['stream_type', 'message_type', 'process_error']
   })
 
+  const pendingEvaluationCounter = MetricsClient.counterWith({})({
+    name: 'ao_process_pending_evaluations',
+    description: 'The total number of pending evaluations on a CU',
+    labelNames: ['type']
+  })
+
+  pendingEvaluationCounter.set({ type: 'primary' }, primaryWorkQueue.size)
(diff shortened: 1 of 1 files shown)


servers/cu/src/bootstrap.js
```diff
@@ -294,6 +294,15 @@ export const createApis = async (ctx) => {
     labelNames: ['stream_type', 'message_type', 'process_error']
   })
 
+  const pendingEvaluationCounter = MetricsClient.counterWith({})({
+    name: 'ao_process_pending_evaluations',
+    description: 'The total number of pending evaluations on a CU',
+    labelNames: ['type']
+  })
+
+  pendingEvaluationCounter.set({ type: 'primary' }, primaryWorkQueue.size)
```
jfrain99, line 303: This will set the metric amount when this file is run at startup, then never again. We will need to move the 'set's to the location that the work queues are incremented (evaluateWith, maybe?) and increment it by 1 on addition and decrement by 1 on completion.

`set` does not currently exist on the grafana counters. Only `inc` is implemented (see effects/metrics.js, counterWith). This can be implemented with inc / dec or feel free to implement set in metrics.js - up to you.

bredamatt: Got it, will adjust this accordingly.

bredamatt: @jfrain99  Seems there is a way to listen to events on the `PQueue` and accordingly `inc`, `dec` the `pendingEvaluationsCounter` given those events take place. I've added this accordingly. Not so familiar with JS, so let me know if I need to change further.

PR #1106, opened 2025-01-10
feat(cu): add option for multiple gql urls for loadblockmeta

jfrain99:
Creates new config, GRAPHQL_URLS, which is an array of GQL urls to be cycled through in the backoff of `loadBlockMeta`


twilson63 (approved):
How often will it retry? looks like it is set to 2, should it be more?

jfrain99:
The actual backoff retries twice, the getResult retries and the total recovery attempts are both new config vars.

PR #1110, opened 2025-01-14
fix(connect): support spawn data ArrayBuffer type

burnto:
Currently aoconnect's `spawn` function's `data` option is incorrectly limited to a string type, whereas the internal implementaiton [supports ArrayBuffer types](https://github.com/permaweb/ao/blob/main/connect/src/lib/spawn/upload-process.js#L27), which is relevant when using non-text data. This small PR fixes this type.


jfrain99:
LGTM!

PR #1113, opened 2025-01-16
feat(cu): implement hash chain validation on messages received from SU

### LINKED ISSUE #1112: feat(cu): implement hash chain validation
For all scheduled messages received from the SU, the CU needs to validate the hash chain. By validating the hash chain, the CU can assert the stream of messages received from the SU are ordered correctly, and does not have any holes.

If the CU's hash validation detects an error in the stream from the SU, the CU will HALT and send a 422 (`Unprocessable Entity`) status code to the client, along with internal logging identifying the message and process id.

TillaTheHun0:
Closes #1112 

This PR implements hash chain validation in the CU as described in #1112.

The implementation accommodates the SU changes portion, to include the `previous_assignment` portion of the response, not yet being rolled out. In this case, the CU skips it's hash chain validation. In other words, the CU's hashchain validation is feature flagged, such to only be turned on when the SU changes are live. This enables flexible deployment of either unit -- the CU before the SU or vice-versa

### DIFF
--- a/servers/cu/src/effects/ao-su.js
+++ b/servers/cu/src/effects/ao-su.js
@@ -1,10 +1,15 @@
 /* eslint-disable camelcase */
 import { Transform, compose as composeStreams } from 'node:stream'
+import { createHash } from 'node:crypto'
+
 import { of } from 'hyper-async'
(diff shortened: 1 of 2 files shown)


TillaTheHun0:
The requirements have changes, and thus requires a different approach. Closing this PR

PR #1118, opened 2025-01-18
Fix[dev-cli]: root_dir: use absolute path instead of relative path

RyanKung:
The output of $(dirname "$0")/.." is a relative path, which may cause an error:

```
zip I/O error: No such file or directory
```


PeterFarber:
@RyanKung 
Thanks for the PR! I agree that absolute paths can remove ambiguity and are more robust!

PR #1120, opened 2025-01-18
feat(ur): enable config to route From-Module

twilson63:


TillaTheHun0 (approved):
LGTM. Nicely done.

I could see a case for caching full SU results, instead of separates bits of the result and reduce calls to the SUs (the current PR will dupe calls to the SUs), but then we'd need to worry about the potential for more cache evictions, which could cause either thrashing or high memory pressure, depending on configuration.

Without further exploration, the approach in the PR -- resulting in dupe calls -- is most likely the safest bet.

TillaTheHun0:
Another approach, probably most ideal, would be to combine the cache entries into a single shared cache, key'd off process id and whose value would contain both the owner and module.

PR #1122, opened 2025-01-19
feat: support building ao wasm with Rust

RyanKung:


RyanKung:
Update, we now have issues:

```rust
[wasm-validator error in function _$LT$$RF$T$u20$as$u20$core..fmt..Debug$GT$::fmt::hea3ded4c73476728] i64 != i32: indirect call target must be an i32, on
(call_indirect $0 (type $func.0)
 (i64.load
  (local.get $0)
 )
 (local.get $1)
 (i64.load offset=24
  (i64.load offset=8
   (local.get $0)
  )
 )
)
```

Looks like `core::fmt::Debug` failed to compile into `i32`.

RyanKung:
> Update, we now have issues:
> 
> ```rust
> [wasm-validator error in function _$LT$$RF$T$u20$as$u20$core..fmt..Debug$GT$::fmt::hea3ded4c73476728] i64 != i32: indirect call target must be an i32, on
> (call_indirect $0 (type $func.0)
>  (i64.load
>   (local.get $0)
>  )
>  (local.get $1)
>  (i64.load offset=24
>   (i64.load offset=8
>    (local.get $0)
>   )
>  )
> )
> ```
> 
> Looks like the call_indirect of `core::fmt::Debug` failed to compile into `i32`.

Fixed by set target=32 in `src/config.yaml`

LcsH0s:
will this be merged any time soon? doesnt seem to be tagged for release

PR #1137, opened 2025-02-05
Vince juliano/mu top up 1127

VinceJuliano:


servers/mu/src/domain/clients/relay.js
```diff
@@ -0,0 +1,61 @@
...
+      url: new URL(urlString),
+      method: 'POST'
+    }
+    
+    const { method, headers } = await signMessage({
+      key: s,
+      fields: [
+        '@path',
+        '@query-param;name=amount',
+        '@query-param;name=recipient',
```
TillaTheHun0, line 39: HyperBEAM doesn't now support this part of RFC9421 (it did but was removed).

Best option would be to add additional headers that contain amount and recipient.

Same with path.

servers/mu/src/domain/lib/write-message-tx.js
```diff
@@ -31,34 +32,59 @@ export function writeMessageTxWith (env) {
...
-        .map(ctxSchemaArweave.parse)
-        .bimap(
-          (e) => {
-            return new Error(e, { cause: { ...ctx, stage: 'write-message' } })
-          },
-          logger.tap({ log: 'Added "arweaveTx" to ctx' })
-        )
-    }
+            if(!amount || !sender) {
+              return Rejected(new Error('Must set Sender and Quantity to top up.', { cause: ctx }))
```
TillaTheHun0, line 44: Will this cause infinite retry in MU? It ought not, or cause a worker thread to crash.

servers/mu/src/domain/clients/relay.js
VinceJuliano: Ok I fixed this in the latest commit, using only headers no derived components

servers/mu/src/domain/lib/write-message-tx.js
VinceJuliano: Made sure there was no infinite looping happening when these fiels are not present

TillaTheHun0 (approved):
LGTM

PR #1151, opened 2025-02-15
fix(loader): stub out getenv

twilson63:


VinceJuliano (commented):
It looks to me like getEnvStrings is now always going to return an empty list but there is other code that uses it and expects it to be returning a list, has this been considered?

image

twilson63:
> It looks to me like getEnvStrings is now always going to return an empty list but there is other code that uses it and expects it to be returning a list, has this been considered?
> 
> image

Yep, it is injecting in wasm memory, which is what we dont want

VinceJuliano (approved):

PR #1156, opened 2025-02-18
feat(connect): add post mainnet mode with relay device

twilson63:


connect/src/client/hb.js
```diff
@@ -222,6 +244,14 @@ export function loadResultWith ({ fetch, logger: _logger, HB_URL, signer }) {
...
 }
 
+export class InsufficientFunds extends Error {
+  name = 'InsufficientFunds'
+}
+
+export class RedirectRequested extends Error {
+  name = 'RedirectRequested'
+}
+
```
TillaTheHun0, line 254: Cool

connect/src/client/hb.js
```diff
@@ -72,14 +72,36 @@ export function requestWith ({ fetch, logger: _logger, HB_URL, signer }) {
...
+                  return {
+                    headers,
+                    body
+                  }
+                } else {
+                  const body = await res.text()
+                  return {
+                    headers: res.headers,
+                    body
+                  }
```
TillaTheHun0, line 100: Any chance that HyperBEAM returns actual bytes? Maybe need a checks for `content-type: application/octet-stream` and then call `res.arrayBuffer()`

connect/src/client/hb.js
```diff
@@ -72,14 +72,36 @@ export function requestWith ({ fetch, logger: _logger, HB_URL, signer }) {
...
         })).then((req) => ({ ...req, body }))
       ))
       .map(logger.tap('Sending HTTP signed message to HB: %o'))
       .chain((request) => of(request)
-        .chain(fromPromise(({ url, method, headers, body }) =>
-          fetch(url, { method, headers, body, redirect: 'follow' })
+        .chain(fromPromise(({ url, method, headers, body }) => {
+          return fetch(url, { method, headers, body, redirect: 'follow' })
+            .then(async res => {
+              if (res.status < 300) {
```
TillaTheHun0, line 84: Nothing is returned if the `res.status` is greater than 300. Should probably be a clause to just return `res`

connect/src/client/hb.js
twilson63: absolutely, but right now it does not return app/json

connect/src/client/hb.js
twilson63: I don't follow everything else returns res

connect/src/client/hb.js
twilson63: maybe in future when we have a way to verify

connect/src/client/ao-mu.js
```diff
@@ -48,7 +48,13 @@ export function deployMessageWith ({ fetch, MU_URL, logger: _logger }) {
               }
             )
           )).bichain(
-            err => Rejected(new Error(`Error while communicating with MU: ${JSON.stringify(err)}`)),
+            err => {
+              if (err.name === 'RedirectRequested') {
+                return Rejected(err)
+              } else {
+                return Rejected(new Error(`Error while communicating with MU: ${JSON.stringify(err)}`))
+              }
```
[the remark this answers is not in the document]
twilson63: prettier likes it like this 🤷

connect/src/client/hb.js
TillaTheHun0: For example, `res.status` is 400, it looks like the `if` check won't match, and then the function ends, which means it returns `undefined`. Shouldn't there be like `else { return res }` or something of the sort?

TillaTheHun0 (approved):
LGTM

PR #1189, opened 2025-03-25
feat(mu): allow custom cu url in push endpoint

jfrain99:


twilson63 (approved):
LGTM

VinceJuliano (commented):
Worried this part might modify the CU_URL for other requests which would then flood the custom url

return async (txId, processId, logId, customCuUrl) => {
    if (customCuUrl) {
      CU_URL = customCuUrl
    }

PR #1195, opened 2025-04-02
feat: create standalone signer / dispatch APIs (signMessage / sendSignedMessage)

NickJ202:
This PR adds two new APIs to aoconnect. The first is a signMessage method which takes the same arguments passed to message, however only returns a signed data item / http message. The second method added is sendSignedMessage, which takes the previously created signed data item / http message and performs the dispatch of the message.


twilson63 (approved):
LGTM

PR #1215, opened 2025-05-12
Introduce unsafe mode

smowden:
Allows passing memory by reference which leads to a big performance improvement.


twilson63 (approved):
LGTM

PR #1241, opened 2025-08-22
fix(su): optimize RocksDB get_latest_message query

VinceJuliano:


servers/su/src/domain/clients/local_store/store.rs
```diff
@@ -855,30 +855,46 @@ impl DataStore for LocalStoreClient {
...
+            
+            if key_str.starts_with(&process_key_prefix) {
+                let assignment_id = String::from_utf8(assignment_id_bytes.to_vec())?;
+                let latest_message = self.get_message(&assignment_id)?;
+                return Ok(Some(latest_message));
+            }
+            
+            // If we've moved past our prefix, no messages exist for this process
+            break;
+        }
```
jyeshe, line 896: as the loop will iterate only once, maybe an `if` would be more explicit for this logic?

jyeshe (commented):
This looks good to me. The `process_key_prefix` is preserved to compare with the first iterator key while the upper bound is used to initialise the iterator.

